PR #2, opened 2025-01-15
Add litellm service

reytchison:


litellm/docker-compose.litellm.yaml
ret2libc, line 1: can we put it in the main compose file?

reytchison: Yep, I was going to do that next.

PR #3, opened 2025-01-15
Adds fuzzer and common queues 

2over12:


compose.yaml
```diff
@@ -1,14 +1,92 @@
...
+      - cache:/data
+      - ./redis:/mount
+    healthcheck:
+      test: [ "CMD-SHELL", "redis-cli ping | grep PONG" ]
+      interval: 1s
+      timeout: 3s
+      retries: 5
+  
+  stimulate-fuzzer-test:
+    image: fuzzer-runnner-image
```
ret2libc, line 35: this needs to be defined somewhere

2over12: i dont get ur question? per compose (afaik) it's defined by the sole container that has a build command for the image (build-bot) and they all share it

PR #9, opened 2025-01-17
orchestrator: create Downloader and TaskRegistry

ret2libc:


common/protos/orchestrator.proto
```diff
@@ -0,0 +1,34 @@
...
+
+message Task {
+    enum TaskType {
+        TASK_TYPE_FULL = 0;
+        TASK_TYPE_DELTA = 1;
+    }
+
+    string message_id = 1;
+    int64 message_time = 2;
+    string task_id = 3;
```
hbrodin, line 13: I made this into it's own type `TaskId` that essentially just wraps the string. I was thinking that'd be helpful to cleanly refer to task id's as opposed to other strings.

common/protos/orchestrator.proto
```diff
@@ -0,0 +1,34 @@
...
+package orchestrator;
+
+message Task {
+    enum TaskType {
+        TASK_TYPE_FULL = 0;
+        TASK_TYPE_DELTA = 1;
+    }
+
+    string message_id = 1;
+    int64 message_time = 2;
```
hbrodin, line 12: Should we wrap the time-fields in a type called `EpochTime` or something? Just to clarify?

orchestrator/src/buttercup/orchestrator/downloader/__cli__.py
```diff
@@ -0,0 +1,73 @@
...
+import requests
+import hashlib
+
+
+def compute_url_sha256(session: requests.Session, url: str) -> str:
+    response = session.get(url, stream=True)
+    response.raise_for_status()
+    sha256_hash = hashlib.sha256()
+    for chunk in response.iter_content(chunk_size=8192):
+        if chunk:
```
hbrodin: Can chunk be None/false?

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+        adapter = HTTPAdapter(
+            pool_connections=100,  # number of connections to keep in pool
+            pool_maxsize=100,  # maximum number of connections in pool
+            max_retries=retry_strategy,
+        )
+        self.session.mount("http://", adapter)
+        self.session.mount("https://", adapter)
+
+    def get_source_type_dir(self, task_id: str, source_type: SourceDetail.SourceType) -> Path:
```
hbrodin: This ties in with the discussion we had yesterday. Basically, there could be multiple sources of the same type. That would cause a collision here.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -0,0 +1,184 @@
...
+        try:
+            # Use session instead of requests directly
+            response = self.session.get(source.url, stream=True)
+            response.raise_for_status()
+
+            # Create directory structure and filename
+            source_dir = self.get_source_type_dir(task_id, source.source_type)
+            # Extract base filename before query parameters
+            filename = source.url.split("/")[-1].split("?")[0]
+            filepath = source_dir / filename
```
hbrodin: I think I'd prefer a hardcoded name or potentially the sha256 or something to ensure that we don't break on this filename processing for whatever reason.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -0,0 +1,184 @@
...
+            logger.info(f"[task {task_id}] Downloading source type {source.source_type} to {filepath}")
+
+            # Download and compute hash simultaneously
+            sha256_hash = hashlib.sha256()
+
+            with open(filepath, "wb") as f:
```
hbrodin: I think we could unify this and `compute_url_sha256` if we created the `HashComputingIterator` (Claude...)
```python
class HashComputingIterator:
    def __init__(self, response):
        self.iterator = response.iter_content(chunk_size=8192)
        self.hash_obj = hashlib.sha256()
    
    def __iter__(self):
        return self
        
    def __next__(self):
        chunk = next(self.iterator)
        if chunk:
            self.hash_obj.update(chunk)
        return chunk
    
    @property
    def hexdigest(self):
        return self.hash_obj.hexdigest()

# Usage:
hash_iterator = HashComputingIterator(response)
with open(filepath, "wb") as f:
    for chunk in hash_iterator:
        f.write(chunk)

computed_hash = hash_iterator.hexdigest
```
Might be handy in other places as well, maybe...

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
hbrodin: Also, I think what you mentioned, having a temporary directory while downloading is probably a good idea.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -0,0 +1,184 @@
...
+        """Process a single task by downloading all its sources"""
+        logger.info(f"Processing task {task.task_id}")
+
+        success = True
+        for source in task.sources:
+            result = self.download_source(task.task_id, source)
+            if result is None:
+                success = False
+
+            if result and (result.suffix.startswith(".tar") or result.name.endswith(".tar.gz")):
```
hbrodin: If the suffix isn't what we expect, I think something is really wrong and we should probably scream loudly... Further, we won't be able to do anything with a partial results (as far as I understand it at least), so I think we could just as well bail when we detect that we failed to download. Or, we'd have to resume downloading of broken files in a next iteration or something. Regardless, I think we could improve the logic a bit here to make it more robust.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -0,0 +1,184 @@
...
+                            for member in tar.getmembers():
+                                if member.name.startswith(first_dir + "/"):
+                                    member.name = member.name[len(first_dir) + 1 :]
+                                    if member.name:  # Only extract if there's a remaining path
+                                        tar.extract(member, path=extract_path)
+                        else:
+                            # If no subdirectory found, extract normally but safely
+                            for member in tar.getmembers():
+                                tar.extract(member, path=extract_path)
+
```
hbrodin: Consider refactor into a separate `extract_archive`-function.

orchestrator/src/buttercup/orchestrator/registry.py
```diff
@@ -0,0 +1,87 @@
...
+    def __contains__(self, task_id: str) -> bool:
+        """Check if a task ID exists in the registry"""
+        return self.redis.hexists(HashNames.TASKS_REGISTRY, task_id.upper())
+
+    def items(self):
+        """Iterate over (task_id, task) pairs in the registry"""
+        for task_id in self:
+            yield task_id, self[task_id]
+
+    def set(self, task_id: str, task: Task):
```
hbrodin: I like that you've wrapped the registry like this. I was thinking how we can ensure that `.upper()` is always called. This helps solve that. However, I'm thinking that we wouldn't need to pass in the `task_id` here as that should already be available from the `task`.
Also, consider implementing a function `prepare_key` or something that is essentially just doing `.upper()`, just to signal that the upper is a way of harmonizing keys regardless of case.

orchestrator/src/buttercup/orchestrator/registry.py
```diff
@@ -0,0 +1,87 @@
...
+        return iter(self.redis.hkeys(HashNames.TASKS_REGISTRY))
+
+    def __contains__(self, task_id: str) -> bool:
+        """Check if a task ID exists in the registry"""
+        return self.redis.hexists(HashNames.TASKS_REGISTRY, task_id.upper())
+
+    def items(self):
+        """Iterate over (task_id, task) pairs in the registry"""
+        for task_id in self:
+            yield task_id, self[task_id]
```
hbrodin: This will require additional roundtrips to redis for each individual task. Consider using `hgetall` instead.

orchestrator/src/buttercup/orchestrator/registry.py
```diff
@@ -0,0 +1,87 @@
+from buttercup.common.datastructures.orchestrator_pb2 import Task
+from redis import Redis
+from buttercup.common.queues import HashNames
+from dataclasses import dataclass
+from collections import UserDict
+
+
+@dataclass
+class TaskRegistry(UserDict):
```
hbrodin: I like this abstraction! However, I'd like to modify the interface a bit to not leak how we represent the task-id in redis. E.g. iteration using `__iter__`, I'd like for that to be the equivalent of a `hgetall` where tasks are deserialized. The id is available from the task in the form or `task.task_id`.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
ret2libc: the tmp dir is implemented in the current version

ret2libc: > This ties in with the discussion we had yesterday. Basically, there could be multiple sources of the same type. That would cause a collision here.

i didn't handle this yet because i asked a question in the slack channel and i'd prefer to have this for now instead of having to deal with different names and more complicated stuff, unless they explicitly say multiple types are allowed.

ret2libc:
let's squash + merge this

PR #11, opened 2025-01-17
common: expand ReliableQueue

ret2libc:
- move protobufs to common package
- group queue names/group names into enums
- create ReliableQueue factory based on msg type
- modify ReliableQueue's code to deal with same-reader pending queue, then new messages, then autoclaim.
- use type hints for reliablequeue
- added `claim_item` with the idea that long-running jobs can claim the items they already have to reset their idle time and avoid other tasks claiming them.

### DIFF
--- a/common/src/buttercup/common/queues.py
+++ b/common/src/buttercup/common/queues.py
@@ -6,7 +6,7 @@ from redis import Redis, RedisError
 from google.protobuf.message import Message
 from buttercup.common.datastructures.fuzzer_msg_pb2 import BuildRequest, BuildOutput
 import logging
-from typing import Type, Generic, TypeVar, overload, Literal
+from typing import Type, Generic, TypeVar
 import uuid
 from enum import Enum
 from typing import Any
@@ -22,6 +22,9 @@ class GroupNames(str, Enum):
     BUILDER_BOT = "build_bot_consumers"
     ORCHESTRATOR = "orchestrator_group"
 
+
+BUILD_TASK_TIMEOUT_MS = 15 * 60 * 1000
+BUILD_OUTPUT_TASK_TIMEOUT_MS = 3 * 60 * 1000
 
 logger = logging.getLogger(__name__)
 
@@ -202,69 +205,30 @@ class ReliableQueue(Generic[MsgType]):
 
 
 @dataclass
-class QueueConfig(Generic[MsgType]):
-    """Configuration for a reliable queue"""
-
-    queue_name: str
-    group_name: str
-    message_type: Type[MsgType]
-
-
 class QueueFactory:
     """Factory for creating common reliable queues"""
 
-    def __init__(self):
-        self.queue_configs: dict[QueueNames, QueueConfig[MsgType]] = {
-            QueueNames.BUILD: QueueConfig(
-                queue_name=QueueNames.BUILD,
-                group_name=GroupNames.BUILDER_BOT,
-                message_type=BuildRequest,
-            ),
-            QueueNames.BUILD_OUTPUT: QueueConfig(
-                queue_name=QueueNames.BUILD_OUTPUT,
-                group_name=GroupNames.ORCHESTRATOR,
-                message_type=BuildOutput,
-            ),
-        }
+    redis: Redis
 
-    @overload
-    def create_queue(
-        self, redis: Redis, queue_name: Literal[QueueNames.BUILD], **kwargs: Any
-    ) -> ReliableQueue[BuildRequest]: ...
+    def create_build_queue(self, **kwargs: Any) -> ReliableQueue[BuildRequest]:
+        return ReliableQueue(
+            queue_name=QueueNames.BUILD,
+            group_name=GroupNames.BUILDER_BOT,
+            redis=self.redis,
+            msg_builder=BuildRequest,
+            task_timeout_ms=BUILD_TASK_TIMEOUT_MS,
+            **kwargs,
+        )
 
-    @overload
-    def create_queue(
-        self, redis: Redis, queue_name: Literal[QueueNames.BUILD_OUTPUT], **kwargs: Any
-    ) -> ReliableQueue[BuildOutput]: ...
-
-    def create_queue(
-        self, redis: Redis, queue_name: QueueNames, **kwargs: Any
-    ) -> ReliableQueue[MsgType]:
-        """
-        Create a reliable queue with predefined configuration, allowing for overrides
-
-        Args:
-            redis: Redis connection
-            queue_name: The name of the queue to create
-            **kwargs: Additional arguments to override default configuration
-        """
-        if queue_name not in self.queue_configs:
-            raise ValueError(f"No configuration found for queue: {queue_name}")
-
-        config = self.queue_configs[queue_name]
-
-        # Start with default configuration
-        queue_args = {
-            "queue_name": config.queue_name,
-            "group_name": config.group_name,
-            "redis": redis,
-            "msg_builder": config.message_type,
-        }
-
-        # Override with any provided kwargs
-        queue_args.update(kwargs)
-
-        return ReliableQueue(**queue_args)
+    def create_build_output_queue(self, **kwargs: Any) -> ReliableQueue[BuildOutput]:
+        return ReliableQueue(
+            queue_name=QueueNames.BUILD_OUTPUT,
+            group_name=GroupNames.ORCHESTRATOR,
+            redis=self.redis,
+            msg_builder=BuildOutput,
+            task_timeout_ms=BUILD_OUTPUT_TASK_TIMEOUT_MS,
+            **kwargs,
+        )
 
 
 @dataclass
--- a/common/tests/test_reliable_queue.py
+++ b/common/tests/test_reliable_queue.py
@@ -8,11 +8,14 @@ from buttercup.common.queues import (
     QueueFactory,
     QueueNames,
     GroupNames,
+    BUILD_TASK_TIMEOUT_MS,
+    BUILD_OUTPUT_TASK_TIMEOUT_MS,
 )
 from buttercup.common.datastructures.fuzzer_msg_pb2 import BuildRequest, BuildOutput
 
 GROUP_NAME = "test_group"
 QUEUE_NAME = "test_queue"
+
 
 @pytest.fixture
 def redis_client():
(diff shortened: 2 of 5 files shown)


ret2libc:
> modify ReliableQueue's code to deal with same-reader pending queue, then new messages, then autoclaim.

code for checking for same-reader pending queue taken from https://redis.io/docs/latest/develop/data-types/streams/

need to think though about the order of these operations. do we want to deal with same-reader pending queue? also, do we want to have them take priority over new messages? There might be a problem where the pending item will always make the reader crash, so the system will enter a loop.

@2over12 i guess that was the reason for not dealing with same-reader pending queue and just use autoclaim IF there are no new messages, right?

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,183 @@ def __init__(self, qname: str, redis: Redis):
...
-            #[[b'fuzzer_build_queue', [(b'1736893605878-0', {b'item': b'\n\x05nginx\x12\tlibfuzzer\x1a\x07address"\x18/home/iansmith/oss-fuzz/'})]]]
-            # we want the first stream, second elem 
-            elem = elem[0][1]
-        if len(elem) <= 0:
-            return None
-        print(elem)
-        print(elem[0][1])
+            stream_item = streams_items[0][1]
+
+        if len(stream_item) == 0:
```
ret2libc:
```suggestion
        if len(stream_item) == 0 and self.last_stream_id != '>':
```

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,183 @@ def __init__(self, qname: str, redis: Redis):
...
+    A single item in a reliable queue.
+    """
+
+    item_id: str
+    deserialized: MsgType
+    consumer_name: str
+
+
+@dataclass
+class ReliableQueue(Generic[MsgType]):
```
ret2libc, line 118: - [x] choose if to check the same-reader pending queue list (0-0)
- [x] check if an item has been in the queue for too long (e.g. something stuck indefinitely). for long-running job, the idea is to claim the job periodically, so to reset the idle time. however, after a very long period of time, we should be able to clean/remove even those tasks (e.g. 2hrs)
- [x] check for how many times we tried to autoclaim an item (this needs to be considered carefully to see how it interacts with long-running job).

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,183 @@ def __init__(self, qname: str, redis: Redis):
...
+                message_type=BuildRequest,
+            ),
+            BuildOutput: QueueConfig(
+                queue_name=QueueNames.BUILD_OUTPUT,
+                group_name=GroupNames.ORCHESTRATOR,
+                message_type=BuildOutput,
+            ),
+        }
+
+    def create_queue(
```
ret2libc: probably the factory should be based on the queue name rather than the msg type. i used the msg type for the typing, but maybe there's another way.

2over12:
> we want to deal with same-reader pending queue? also, do we want to have them take priority over new messages? There might be a problem where the pendi

I dont know what you are saying with "same-reader pending queue" but yes that was the reason for just autoclaiming when a worker has nothing else to do. If there is not a new item to read on the stream then the idea is we can do no better than claiming it again

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,183 @@ def __init__(self, qname: str, redis: Redis):
...
-        if len(elem) <= 0:
-            return None
-        print(elem)
-        print(elem[0][1])
+            stream_item = streams_items[0][1]
+
+        if len(stream_item) == 0:
+            # No message found in the pending queue for this reader
+            # Go to the new messages
+            self.last_stream_id = ">"
```
2over12, line 183: this is nasty best i can tell what you are trying to do here is: when the queue is first constructed we try to pull from pending, then flip to reading new messages. I think this is a bit overly complex? if we always register new consumer group names on crash we will just xautoclaim from workers that are now dead? i dont see a reason to risk seeing messages twice e.g. by constructing a queue twice etc, this makes qs oddly stateful

ret2libc:
> > we want to deal with same-reader pending queue? also, do we want to have them take priority over new messages? There might be a problem where the pendi
> 
> I dont know what you are saying with "same-reader pending queue" but yes that was the reason for just autoclaiming when a worker has nothing else to do. If there is not a new item to read on the stream then the idea is we can do no better than claiming it again

From https://redis.io/docs/latest/develop/data-types/streams/#creating-a-consumer-group 

> basically [XREADGROUP](https://redis.io/docs/latest/commands/xreadgroup/) has the following behavior based on the ID we specify:

> If the ID is the special ID > then the command will return only new messages never delivered to other consumers so far, and as a side effect, will update the consumer group's last ID.
> If the ID is any other valid numerical ID, then the command will let us access our history of pending messages. That is, the set of messages that were delivered to this specified consumer (identified by the provided name), and never acknowledged so far with [XACK](https://redis.io/docs/latest/commands/xack/).

The code, as said, was taken from the same page a little bit below. Anyway, I'm fine to remove that logic with the idea that autoclaim will always do that eventually.

common/src/buttercup/common/queues.py
ret2libc: See issue https://github.com/trailofbits/afc-crs-trail-of-bits/issues/14

common/src/buttercup/common/queues.py
ret2libc: ok, fixed by making this behaviour non-default. it's still possible, but it won't happen unless user creates the ReliableQueue with `last_stream_id='0-0'`.

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,194 @@ def __init__(self, qname: str, redis: Redis):
     def __iter__(self):
         return iter(self.redis.lrange(self.qname, 0, -1))
 
+
```
hbrodin, line 84: Are we using the `NormalQueue` and related structures?

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,194 @@ def __init__(self, qname: str, redis: Redis):
...
+class ReliableQueue(Generic[MsgType]):
+    """
+    A queue that is reliable and can be used to process tasks in a distributed environment.
+    """
+
+    queue_name: str
+    group_name: str
+    redis: Redis
+    msg_builder: Type[MsgType]
+    task_timeout_ms: int = 180000
```
hbrodin, line 127: This would by default claim messages that have been pending for more than three minutes. I think that might be a bit low. Especially if we consider builds that might run for quite some time.

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,194 @@ def __init__(self, qname: str, redis: Redis):
...
+        self.redis.xack(self.queue_name, self.group_name, rq_item.item_id)
+
+    def claim_item(self, item_id: str, min_idle_time: int = 0) -> None:
+        self.redis.xclaim(
+            self.queue_name, self.group_name, self.reader_name, min_idle_time, [item_id]
+        )
+
+
+@dataclass
+class QueueConfig(Generic[MsgType]):
```
hbrodin: I think the timeout for a stream should be part of this configuration as it is inherently tied to the purpose of the queue.

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,194 @@ def __init__(self, qname: str, redis: Redis):
...
-    
-    def ack_item(self, rq_item: RQItem):
-        self.redis.xack(self.qname, self.gname, rq_item.item_id)
+        msg.ParseFromString(message_data[self.INAME])
+
+        return RQItem[MsgType](
+            item_id=message_id, deserialized=msg, consumer_name=self.reader_name
+        )
+
+    def ack_item(self, rq_item: RQItem[MsgType]) -> None:
```
hbrodin: I think we should id the items the same way in `ack_item` and `claim_item`. Consider either replace the `rq_item` with an `item_id` string or modify the `claim_item` to accept an `RQItem`. I'd personally prefer the former.

common/src/buttercup/common/queues.py
```diff
@@ -65,74 +78,194 @@ def __init__(self, qname: str, redis: Redis):
...
+    ) -> ReliableQueue[BuildRequest]: ...
+
+    @overload
+    def create_queue(
+        self, redis: Redis, queue_name: Literal[QueueNames.BUILD_OUTPUT], **kwargs: Any
+    ) -> ReliableQueue[BuildOutput]: ...
+
+    def create_queue(
+        self, redis: Redis, queue_name: QueueNames, **kwargs: Any
+    ) -> ReliableQueue[MsgType]:
```
hbrodin: If we are doing this, couldn't we just as well just have the explicit functions:
`create_build_queue()` and `create_build_output_queue()`? I.e. no need to pass the queue name as string, just call the associated function. The functions could obviously call `create_queue` with the corresponding  name internally.

common/tests/test_reliable_queue.py
```diff
@@ -0,0 +1,207 @@
...
+        reader_name="test_reader2",
+    )
+    item = queue.pop()
+    assert item is not None
+    assert item.deserialized.fields["key"].string_value != "value_0"
+
+    # Ack the message
+    queue.ack_item(item)
+
+    # Simulaten another restart and a new reader
```
hbrodin:
```suggestion
    # Simulate another restart and a new reader
```

common/tests/test_reliable_queue.py
```diff
@@ -0,0 +1,207 @@
...
+    msg_id = item.item_id
+    assert item.deserialized.fields["test_key"].string_value == "test_value_0"
+    # No ack, so pending
+
+    time.sleep(2)
+
+    # Simulate another reader
+    queue = ReliableQueue[Struct](
+        queue_name="test_queue",
+        group_name="test_group",
```
hbrodin: consider extracting constants for `test_queue` and `test_group`.

common/src/buttercup/common/queues.py
ret2libc: The fuzzer is using those.

common/src/buttercup/common/queues.py
ret2libc: Good point! Will update

common/src/buttercup/common/queues.py
ret2libc: those overloads above are just for the IDE tbh, so they can properly recognize the types of the deserialized items. I'm not a big fan of having different methods for each queue (it defeats a bit the purpose of the factory). But I also agree it is not ideal to have the overload for each type... Will think a bit more about this.

common/src/buttercup/common/queues.py
ret2libc: i think this is fine and we should have defaults for each common queue type

PR #19, opened 2025-01-21
Update redis ports

reytchison:


2over12:
im confused by why this change is desired? we dont want to assume the redis url is local etc and you can connect to it via the docker hostname etc etc

reytchison:
For security, I wanted to only publish the ports to the local machine, instead of the default of publishing them to `0.0.0.0`.

PR #21, opened 2025-01-22
Fix CI

ret2libc:


ret2libc:
@2over12 i've done very small modifications to the fuzzer, just to have an initial good green base for all components!

fuzzer/tests/test_fuzzer.py
```diff
@@ -0,0 +1,11 @@
+from buttercup.fuzzing_infra.runner import Runner, Conf
```
2over12: What is this test?

2over12: basically if it's intended to be a stub make it a stub right now it claims it's testing for something it isnt

PR #25, opened 2025-01-24
orchestrator: introduce the "scheduler" component

ret2libc:
the `buttercup-scheduler` component will move some small messages around and do a minimal amount of work on them, like sending the build requests to the fuzzer/code-indexer or setting up the fuzzer weighted map, etc.

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
+++ b/orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
@@ -50,6 +50,24 @@ class Scheduler:
 
         raise RuntimeError(f"Couldn't handle task {task.task_id}")
 
+    def serve_ready_task(self) -> bool:
+        """Handle a ready task"""
+        task_ready_item: RQItem[TaskReady] | None = self.ready_queue.pop()
+
+        if task_ready_item is not None:
+            task_ready: TaskReady = task_ready_item.deserialized
+            try:
+                build_request = self.process_ready_task(task_ready.task)
+                self.build_requests_queue.push(build_request)
+                self.ready_queue.ack_item(task_ready_item.item_id)
+                logger.info(f"Pushed build request for task {task_ready.task.task_id} to build requests queue")
+                return True
+            except Exception as e:
+                logger.error(f"Failed to process task {task_ready.task.task_id}: {e}")
+                return False
+
+        return False
+
     def serve(self):
         """Main loop to process tasks from queue"""
         if self.ready_queue is None:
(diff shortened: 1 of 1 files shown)


orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
```diff
@@ -0,0 +1,82 @@
...
+        while True:
+            task_ready_item: RQItem[TaskReady] | None = self.ready_queue.pop()
+
+            if task_ready_item is not None:
+                task_ready: TaskReady = task_ready_item.deserialized
+                try:
+                    build_request = self.process_ready_task(task_ready.task)
+                except Exception as e:
+                    logger.error(f"Failed to process task {task_ready.task.task_id}: {e}")
+                    continue
```
hbrodin: Wouldn't this prevent the overall progress of the loop? Doesn't matter right now but as we add more scheduler logic I don't think we want this. Consider shifting the build_requests_queue push and ack into the try block instead.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -164,6 +166,9 @@ def serve(self):
         if self.task_queue is None:
             raise ValueError("Task queue is not initialized")
 
+        if self.ready_queue is None:
+            raise ValueError("Ready queue is not initialized")
+
```
hbrodin, line 171: Ideally, we wouldn't even get here if we don't have task queues. Could we raise exception in __post_init__ instead?

orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
```diff
@@ -0,0 +1,82 @@
...
+
+        raise RuntimeError(f"Couldn't handle task {task.task_id}")
+
+    def serve(self):
+        """Main loop to process tasks from queue"""
+        if self.ready_queue is None:
+            raise ValueError("Ready queue is not initialized")
+
+        if self.build_requests_queue is None:
+            raise ValueError("Build requests queue is not initialized")
```
hbrodin, line 77: Same comment as for the downloader, can we prevent even reaching here?

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
ret2libc: the `Downloader` can also be instantiated without a redis client (in case you want to do a one-shot run with `buttercup-task-downloader process ...` instead of the whole redis thing. In that case, you can't call `serve` on the `Downloader` object.

orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
ret2libc: Same as the downloader, the `Scheduler` ideally can also be called with a `process` subcommand that just does some operations without the whole redis interactions. So in that case, you instantiate `Scheduler` without a redis client and then you can't call `serve`.

PR #26, opened 2025-01-24
docker-compose: disable redis persistence during development

ret2libc:
What do you think @2over12 ? I feel it's just easier during development to turn off/on the redis server to start from scratch.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -21,10 +21,12 @@ services:
     restart: always
     ports:
       - '127.0.0.1:6379:6379'
-    command: redis-server /mount/redis.conf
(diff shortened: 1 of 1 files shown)


2over12:
probably fine id probably prefer enabling the redis configuration mount so we test with our redis config and can modify and maybe just keeping the actual cache dir is non persistent but im not gonna block on something like that

reytchison:
I would like to disable persistence for now, so I can have empty queues each time I run `docker compose up`.

ret2libc:
Happy to re-evaluate or find a better alternative soon!

PR #28, opened 2025-01-24
fuzzer: add a few logs and some basic settings

ret2libc:


compose.yaml
```diff
@@ -47,23 +47,26 @@ services:
     build:
       context: ./
       dockerfile: ./fuzzer/dockerfiles/runner_image.Dockerfile
-    command: -m buttercup.fuzzing_infra.orchestrator --redis_url redis://redis:6379
+    command: /fuzzer/.venv/bin/python -m buttercup.fuzzing_infra.orchestrator --redis_url redis://redis:6379
```
ret2libc, line 50: these are going to be removed from the compose soon, because not part of the "CRS", so i didn't bother changing these.

2over12: im not sure there's a downside to having a testing profile that boots a fuzz _test e.g. fuzz_test, that runs it in isolation

ret2libc: we could keep it, but soon these things are going to be done by the real orchestrator (see #25 ). I agree though we need a nice way to "inject" messages in the middle of the system. We could have a simple utility that just connect to redis and inject whatever message you want in a queue you select. So with that we could start the crs workflow from a given point without the previous steps.

PR #31, opened 2025-01-26
Adds source path to BuildRequest

2over12:
open question, are we worried about races on the source dir itself since it is mounted for the build? I think probably

### DIFF
--- a/fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
+++ b/fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
@@ -54,7 +54,7 @@ def main():
                 source_snapshot = os.path.join(wdirstr, f"source-snapshot-{dirid}-{msg.package_name}")
                 shutil.copytree(msg.source_path, source_snapshot)
                 source_path_output = source_snapshot
-            
+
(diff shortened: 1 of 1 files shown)


2over12:
Added snapshot of source dir to prevent race condition, it may be better if we package a challenge as:
```
chal1:
/tooling
/source_dir
```
or something then the build request only needs to point to the challenge and source dir and tooling dir are already defined. Then snapshotting only involves the challenge dir

2over12:
> Isn't it better to have the `task_id` passed in the message instead of a path? That way we 1) know what is the task associated with a given build request 2) can delete the msg in the queue in case the task gets deleted 3) we can determine the path by doing CRS_SCRATCH_DIR/.
> 
> What do you think?

This doesnt quite work because the builder can build multiple versions per task ID (e.g. different engine or sanitizer). My current thinking is just to add a task ID provenance value to everything that is primarily used for deletions. The idea with structuring stuff like pointing to a harness etc is to ensure we can split off resources e.g. for multiple fuzzers on a task etc

2over12:
Another way to say it is we have fan out on a task so a task identifier has multiple build requests and fuzzing items so a task id is only provenance and not uniquely identifying

PR #33, opened 2025-01-26
Fuzzer idempotent map

### LINKED ISSUE #16: Make the target set list idempotent
Right now we are representing the Target weights as just a list. This causes duplicate harnesses if multiple builders finish a target (e.g. xautoclaims from another builder then the other builder finishes). I think it should be safe to hash something like the encoding of:
```
message BuildRequest {
    string package_name = 1;
    string engine = 2;
    string sanitizer = 3;
    string ossfuzz = 4;
}
```

to dedup targets

2over12:
Closes #16

### DIFF
--- a/fuzzer/src/buttercup/fuzzing_infra/fuzzer_bot.py
+++ b/fuzzer/src/buttercup/fuzzing_infra/fuzzer_bot.py
@@ -12,6 +12,7 @@ import tempfile
 import distutils
 from buttercup.common.logger import setup_logging
 from redis import Redis
+
 logger = setup_logging(__name__)
 
 


common/src/buttercup/common/queues.py
```diff
@@ -258,3 +257,4 @@ class BuildConfiguration:
     project_id: str
     engine: str
     sanitizer: str
+    source_path: str | None
```
ret2libc: are these actually `common` stuff? shall they get moved in the fuzzer?

common/uv.lock
ret2libc, line 1: do you know why the update the `common/uv.lock`? And why it adds `pymongo`?

fuzzer/uv.lock
ret2libc, line 1: same question here re uv.lock

common/src/buttercup/common/maps.py
```diff
@@ -0,0 +1,54 @@
+from typing import Generic, TypeVar, Type, Iterator
+from buttercup.common.datastructures.fuzzer_msg_pb2 import WeightedTarget
+import redis
+from bson.json_util import dumps, CANONICAL_JSON_OPTIONS
+
+
+MsgType = TypeVar("MsgType")
```
ret2libc:
```suggestion
MsgType = TypeVar("MsgType", bound=Message)
```

common/src/buttercup/common/maps.py
```diff
@@ -0,0 +1,54 @@
...
+MSG_FIELD_NAME = b"msg"
+
+
+class RedisMap(Generic[MsgType]):
+    def __init__(self, redis_url: str, hash_name: str, msg_builder: Type[MsgType]):
+        self.redis_url = redis_url
+        self.redis = redis.Redis.from_url(redis_url)
```
ret2libc:
```suggestion
    def __init__(self, redis: Redis, hash_name: str, msg_builder: Type[MsgType]):
        self.redis = redis
```

common/src/buttercup/common/maps.py
```diff
@@ -0,0 +1,54 @@
...
+from buttercup.common.datastructures.fuzzer_msg_pb2 import WeightedTarget
+import redis
+from bson.json_util import dumps, CANONICAL_JSON_OPTIONS
+
+
+MsgType = TypeVar("MsgType")
+MSG_FIELD_NAME = b"msg"
+
+
+class RedisMap(Generic[MsgType]):
```
ret2libc, line 11: `@dataclass` is more declarative imo

common/src/buttercup/common/maps.py
```diff
@@ -0,0 +1,54 @@
...
+        for key in self.redis.hkeys(self.hash_name):
+            yield self.get(key)
+
+
+FUZZER_MAP_NAME = "fuzzer_target_list"
+
+
+class FuzzerMap:
+    def __init__(self, redis_url: str):
+        self.mp: RedisMap[WeightedTarget] = RedisMap(redis_url, FUZZER_MAP_NAME, WeightedTarget)
```
ret2libc:
```suggestion
    def __init__(self, redis: Redis):
        self.mp: RedisMap[WeightedTarget] = RedisMap(redis, FUZZER_MAP_NAME, WeightedTarget)
```

common/src/buttercup/common/maps.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __iter__(self) -> Iterator[MsgType]:
+        for key in self.redis.hkeys(self.hash_name):
+            yield self.get(key)
+
+
+FUZZER_MAP_NAME = "fuzzer_target_list"
+
+
+class FuzzerMap:
```
ret2libc, line 37: shouldn't this be a subclass of `RedisMap`?

common/src/buttercup/common/maps.py
ret2libc, line 11: i'm going to reuse this in the orchestrator as well. See https://github.com/trailofbits/afc-crs-trail-of-bits/blob/main/orchestrator/src/buttercup/orchestrator/registry.py .

I used hgetall because it was mainly for debugging and don't expected to have many tasks at all, but i guess your `__iter__` approach is better when you might have more/bigger items in the set.

common/uv.lock
ret2libc: oh i see you are adding ` "pymongo>=4.10.1",` to the deps... why is that?

common/src/buttercup/common/maps.py
2over12: No it hides map internals

common/uv.lock
2over12: canonical json serialization of the key for the map... sorry. https://pymongo.readthedocs.io/en/stable/api/bson/json_util.html#bson.json_util.CANONICAL_JSON_OPTIONS im not sure this is strictly required because we could just use a json ordered list, enforce spacing etc but this felt safer idk

2over12: we need serialization of the key to be canonical so we actually dedup the task

2over12: to be clear here bson that doesnt depend on mongo doesnt have the canonical option hence here we are

PR #34, opened 2025-01-26
Separate corpuses and Hash Inputs

### LINKED ISSUE #8: Partition corpuses by harness
Currently when fuzzing we add a directory buttercup_corpus, but this is not deduped by a harness id so we will point to the same project and accidentally mix corpuses for harnesses together

2over12:
Closes #8 also makes sure that inputs are named by hash, this avoids interference between parallel fuzzers by copying fresh names. TODO could use a faster hash

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -110,7 +110,7 @@ services:
     build:
       context: ./
       dockerfile: ./fuzzer/dockerfiles/runner_image.Dockerfile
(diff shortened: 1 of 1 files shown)


common/src/buttercup/common/corpus.py
ret2libc, line 1: do you see this being used outside of the fuzzer package? seed-gen maybe?

2over12: i put it here for seed-gen to use it to copy fresh inputs into the directory

PR #37, opened 2025-01-27
Delete task implementation

hbrodin:
Implements the HTTP DELETE /v1/task/taskid endpoint and associated functionality.
Note: The intention is that this will be merged with the scheduler once we have that in. It is not supposed to be a free-standing cli/container.

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/registry.py
+++ b/orchestrator/src/buttercup/orchestrator/registry.py
@@ -1,8 +1,7 @@
-import time
 from buttercup.common.datastructures.orchestrator_pb2 import Task
 from redis import Redis
 from buttercup.common.queues import HashNames
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from google.protobuf import text_format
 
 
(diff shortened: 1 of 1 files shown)


orchestrator/src/buttercup/orchestrator/cancellation/cancellation.py
```diff
@@ -0,0 +1,130 @@
...
+
+    sleep_time: float = 0.1
+    redis: Redis | None = None
+    delete_queue: ReliableQueue | None = field(init=False, default=None)
+    registry: TaskRegistry | None = field(init=False, default=None)
+
+    def __post_init__(self):
+        """Initialize Redis connection, deletion queue and task registry."""
+        if self.redis is None:
+            self.redis = get_redis()
```
ret2libc, line 42: `get_redis` is actually for the task server and it uses the TaskServerSettings, so it should not be used outside of that component.

orchestrator/src/buttercup/orchestrator/cancellation/cancellation.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    logging.basicConfig(level=logging.INFO)
+    logger.info("Starting cancellation service")
+
+    cancellation = Cancellation()
+    cancellation.run()
+
+
+if __name__ == "__main__":
+    main()
```
ret2libc, line 130: can this be removed? there's already __cli__.py which does similar things

orchestrator/src/buttercup/orchestrator/cancellation/config.py
ret2libc, line 1: there are some dupped Settings for this component. Either put it in here or in __cli__.py

hbrodin:
Closed as #39 replaces this one.

PR #38, opened 2025-01-27
orchestrator/scheduler: process build output messages for the fuzzer

ret2libc:


orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
```diff
@@ -37,19 +43,30 @@ def mock_process_ready_task(self, task: Task) -> BuildRequest:
...
             return self.mock_process_ready_task(task)
 
         raise RuntimeError(f"Couldn't handle task {task.task_id}")
 
+    def process_build_output(self, build_output: BuildOutput) -> list[WeightedTarget]:
+        """Process a build output"""
+        logger.info(
+            f"Processing build output for {build_output.package_name}/{build_output.engine}/{build_output.sanitizer}/{build_output.output_ossfuzz_path}"
+        )
+        build_dir = Path(build_output.output_ossfuzz_path) / "build" / "out" / build_output.package_name
```
hbrodin, line 73: Consider refactoring the path generation to a separate function/structure. I think we'll need to be able to interact with these paths in several places.

orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
```diff
@@ -68,19 +85,38 @@ def serve_ready_task(self) -> bool:
...
+
     def serve(self):
         """Main loop to process tasks from queue"""
-        if self.ready_queue is None:
-            raise ValueError("Ready queue is not initialized")
-
-        if self.build_requests_queue is None:
-            raise ValueError("Build requests queue is not initialized")
+        if self.ready_queue is None or self.build_output_queue is None or self.fuzzer_map is None:
+            raise ValueError("Queue or map is not initialized")
```
hbrodin: Consider rephrasing to indicate more specifically what is wrong or should be done from the user.

orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
ret2libc: I prefer thinking about these extractions once I see at least the same pattern happening twice!

PR #41, opened 2025-01-28
common: redesign the QueueFactory for flexible queue creation

ret2libc:
I think the initial approach of having just a single `create_queue` method is better than having X methods because you can more easily create a custom queue.

For example, I really want to have a "message dispatcher" utility that can be used to inject custom messages in one of the queues, so that you can test one of the components without following the whole flow.

Moreover, while thinking about it, I realized that we are now defaulting to a single consumer group name, but this is risky. Each consumer should explicitly set its consumer name. Right now we were re-using the consumer name of another group in order to push messages in a queue, however you don't need any consumer group name to just push stuff (this is common for outputs of components, that want to publish a message in another queue).

So, to make things clearer:
- if you just want to push stuff in a queue
```python
q = QueueFactory(redis).create_queue(QueueNames.BUILD_REQUESTS)
q.push(item)
```
- if you want to read/write from/to a queue. Note that if you don't provide the GroupName, the queue won't be allowed to pop items. This explicit group name, prevents us from "stealing" messages meant for some other groups
```python
# fuzzer bot
q = QueueFactory(redis).create_queue(QueueNames.BUILD_OUTPUTS, GroupNames.FUZZER_BOT)
item = q.pop()

...

# contextualizer
q = QueueFactory(redis).create_queue(QueueNames.BUILD_OUTPUTS, GroupNames.CONTEXTUALIZER)
item = q.pop()
```
- if you want to read from a queue, but you don't define a consumer name, an exception is raised
```python
q = QueueFactory(redis).create_queue(QueueNames.BUILD_OUTPUTS)
item = q.pop() # raise ValueError()
```

### DIFF
--- a/common/tests/test_reliable_queue.py
+++ b/common/tests/test_reliable_queue.py
@@ -219,6 +219,7 @@ def test_queue_factory(redis_client):
     assert item.deserialized.ossfuzz == "test_ossfuzz"
     queue.ack_item(item.item_id)
 
+
 def test_invalid_group_name():
     factory = QueueFactory(redis_client)
     with pytest.raises(ValueError):


hbrodin:
I was just thinking at the same issue with group names when I saw this. Nice job!

common/src/buttercup/common/queues.py
```diff
@@ -206,58 +216,95 @@ def pop(self) -> RQItem[MsgType] | None:
...
+            "redis": self.redis,
+            "queue_name": config.queue_name,
+            "msg_builder": config.msg_builder,
+            "task_timeout_ms": config.task_timeout_ms,
+        }
+        if group_name is not None and group_name in config.group_names:
+            queue_args["group_name"] = group_name
+
+        queue_args.update(kwargs)
+        return ReliableQueue(**queue_args)
```
hbrodin: I think it would be nice to prevent calling pop on a `ReliableQueue` that does not have a group name set? Can we raise an exception? Or is there ever a case when we want that?

common/src/buttercup/common/queues.py
```diff
@@ -206,58 +216,95 @@ def pop(self) -> RQItem[MsgType] | None:
...
+    def create_queue(
+        self, queue_name: Literal[QueueNames.DOWNLOAD_TASKS], group_name: GroupNames, **kwargs: Any
+    ) -> ReliableQueue[TaskDownload]: ...
+
+    @overload
+    def create_queue(
+        self, queue_name: Literal[QueueNames.READY_TASKS], group_name: GroupNames, **kwargs: Any
+    ) -> ReliableQueue[TaskReady]: ...
+
+    def create_queue(
```
hbrodin: nit: We could rename this to just create or something as it is the `QueueFactory`.

common/src/buttercup/common/queues.py
hbrodin: never mind. Just saw that you had already wrapped the pop function...

PR #43, opened 2025-01-28
patcher: base skeleton

ret2libc:


common/src/buttercup/common/queues.py
```diff
@@ -302,6 +306,12 @@ class QueueFactory:
                 DELETE_TASK_TIMEOUT_MS,
                 [GroupNames.SCHEDULER_DELETE_TASK],
             ),
+            QueueNames.PATCHES: QueueConfig(
+                QueueNames.PATCHES,
+                Patch,
+                PATCH_TASK_TIMEOUT_MS,
+                [],
```
hbrodin, line 313: Guess we will need to change this... Maybe we can just use `GropuNames.ORCHESTRATOR`?

patcher/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ADD common /app/common
+ADD patcher /app/patcher
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    cd patcher && uv sync --frozen --no-editable
+
+
+FROM python:${PYTHON_BASE} AS runtime
+
+COPY --from=builder --chown=app:app /app/patcher/.venv /app/patcher/.venv
```
ret2libc, line 30: nope, this is used to copy dirs from the builder stage into the runtime stage.

common/src/buttercup/common/queues.py
ret2libc: yeah tbh i want to rename all the group names because i think we messed up a bit there. it would be simpler to have group names ~= component (e.g. scheduler, fuzzer, builder, patcher, etc.)

i didn't put anything here yet as there is no user for this.

PR #49, opened 2025-01-29
Primitive vulnerability processing

hbrodin:


orchestrator/src/buttercup/orchestrator/scheduler/vulnerabilities.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def submit_vulnerability(self, crash: Crash) -> None:
+        """
+        Submit the vulnerability to the confirmed vulnerabilities queue
+        """
+        logger.info(f"Submitting confirmed vulnerability for crash in {crash.target.package_name}")
+        # TODO: This is where we would submit the vulnerability to the competition api and get the vuln_id back
+        confirmed_vuln = ConfirmedVulnerability()
+        confirmed_vuln.crash.CopyFrom(crash)
+        confirmed_vuln.vuln_id = str(uuid.uuid4())  # Generate a unique ID for the vulnerability
```
ret2libc:
```suggestion
        confirmed_vuln.vuln_id = str(uuid.uuid4())  # TODO: ID got from the Competition API
```

orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
```diff
@@ -129,15 +132,14 @@ def serve(self):
...
-        since this suggests more work may be available. If no work was done, the loop
-        sleeps briefly to reduce system load.
+        3. Process crashes and vulnerabilities from queues
         """
         if self.redis is None:
             raise ValueError("Redis is not initialized")
 
+        if self.vulnerabilities is None:
+            raise ValueError("Vulnerabilities service is not initialized")
+
```
ret2libc: I think you can remove this, i've replaced all those checks with just `self.redis is None` because those other fields can be None only if `Redis` is none. If something fails during those objects' creations, they would just raise an exception.

orchestrator/src/buttercup/orchestrator/scheduler/vulnerabilities.py
```diff
@@ -0,0 +1,81 @@
...
+    redis: Redis
+    sleep_time: float = 1.0
+    crash_queue: ReliableQueue = field(init=False)
+    unique_vulnerabilities_queue: ReliableQueue = field(init=False)
+    confirmed_vulnerabilities_queue: ReliableQueue = field(init=False)
+
+    def __post_init__(self):
+        queue_factory = QueueFactory(self.redis)
+        self.crash_queue = queue_factory.create(QueueNames.CRASH, GroupNames.ORCHESTRATOR, block_time=None)
+        self.unique_vulnerabilities_queue = queue_factory.create(QueueNames.UNIQUE_VULNERABILITIES, block_time=None)
```
ret2libc: this needs a group name

PR #52, opened 2025-01-29
orchestrator/downloader: apply diff if DELTA task

ret2libc:


orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -143,6 +186,14 @@ def process_task(self, task: Task) -> bool:
                         break
 
                     source.path = extracted_dir
+                    source_paths[source.source_type] = tmp_task_dir / extracted_dir
```
hbrodin: Right now it is only one per source type but it seems we are required to support multiple repos per source type. I'd suggest at least making a comment here that we will need to handle that.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -143,6 +186,14 @@ def process_task(self, task: Task) -> bool:
...
+
+            # If this is a delta task, apply the diff to the source code
+            if success and task.task_type == Task.TaskType.TASK_TYPE_DELTA:
+                logger.info(f"[task {task.task_id}] Applying diff to source code")
+                diff_dir = source_paths.get(SourceDetail.SourceType.SOURCE_TYPE_DIFF)
+                assert diff_dir is not None, "Missing diff source for delta task"
```
hbrodin: Idk, assert doesn't feel like the right tool here. I'd consider either raising an exception or signaling that something is off in another way. I know assert will raise an assertionerror but still...

orchestrator/Dockerfile
```diff
@@ -27,6 +27,11 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 
 FROM python:${PYTHON_BASE} AS runtime
 
+# Install required system packages
+RUN apt-get update && \
+    apt-get install -y patch && \
```
hbrodin, line 32: might want to add `DEBIAN_FRONTEND=noninteractive ` as well.

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -119,6 +120,46 @@ def safe_extract(tar, path: Path) -> None:
...
+            # Find the first .patch or .diff file in the directory
+            diff_files = list(diff_dir.glob("*.patch")) + list(diff_dir.glob("*.diff"))
+            if not diff_files:
+                # If no .patch or .diff files found, try any file
+                diff_files = list(diff_dir.glob("*"))
+
+            if not diff_files:
+                raise FileNotFoundError("No diff file found in the extracted directory")
+
+            diff_file = diff_files[0]
```
hbrodin: Would it make sense to have a loop over diff_files?

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -119,6 +120,46 @@ def safe_extract(tar, path: Path) -> None:
...
+            logger.error(f"[task {task_id}] Error applying diff: {str(e)}")
+            logger.debug(f"[task {task_id}] Error returncode: {e.returncode}")
+            logger.debug(f"[task {task_id}] Error stdout: {e.stdout}")
+            logger.debug(f"[task {task_id}] Error stderr: {e.stderr}")
+            return False
+        except Exception as e:
+            logger.error(f"[task {task_id}] Error applying diff: {str(e)}")
+            return False
+
     def process_task(self, task: Task) -> bool:
```
hbrodin, line 216: I'd consider starting to break this function into smaller pieces. Not critical yet.

PR #53, opened 2025-01-30
Merge protobufs

hbrodin:
The two protobufs fuzzer_msg.proto and orchestrator.proto has been replaced with a single msg.proto. This is to make it easier to embedd structures in other structures across protos. Just compoling the protos work well, but python imports break. As there aren't that many structures merging seamed like an easy fix.

### DIFF
--- a/fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
+++ b/fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
@@ -4,7 +4,7 @@ from redis import Redis
 import argparse
 import tempfile
 from buttercup.common.queues import RQItem, QueueFactory
-from buttercup.common.datastructures.fuzzer_msg_pb2 import BuildRequest, BuildOutput
+from buttercup.common.datastructures.msg_pb2 import BuildRequest, BuildOutput
 from buttercup.common.logger import setup_logging
 import shutil
 import time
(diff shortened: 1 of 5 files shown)


common/protos/msg.proto
```diff
@@ -1,7 +1,8 @@
 syntax = "proto3";
 
-package orchestrator;
+package msgpb2;
 
+// Task-related messages (from orchestrator.proto)
```
ret2libc: i'd remove the "from *.proto" because those files won't exist anymore.

orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
```diff
@@ -5,9 +5,9 @@
 from redis import Redis
 from buttercup.common.queues import ReliableQueue, QueueFactory, RQItem, QueueNames, GroupNames
 from buttercup.common.maps import FuzzerMap
-from buttercup.common.datastructures.orchestrator_pb2 import TaskReady, Task, SourceDetail
+from buttercup.common.datastructures.msg_pb2 import TaskReady, Task, SourceDetail
 from buttercup.orchestrator.scheduler.cancellation import Cancellation
-from buttercup.common.datastructures.fuzzer_msg_pb2 import BuildRequest, BuildOutput, WeightedTarget
+from buttercup.common.datastructures.msg_pb2 import BuildRequest, BuildOutput, WeightedTarget
```
ret2libc: merge this with the previous import

PR #56, opened 2025-01-31
Early vulnerability submission implementation

### LINKED ISSUE #55: Integrate the vulnerability processing with the competition api
- [x] Invoke the api with a deduped vuln.
- [x] Spin up the docker container api mock and ensure it works correctly

hbrodin:
This doesn't read the actual bytes that trigger. Further, it has no idea what the taskid is as that is not yet available for crashes.

Closes #55

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -165,10 +165,10 @@ services:
       - BUTTERCUP_SCHEDULER_SERVE__REDIS_URL=redis://redis:6379
       - BUTTERCUP_SCHEDULER_LOG_LEVEL=debug
       - BUTTERCUP_SCHEDULER_SERVE__SLEEP_TIME=5
+      - BUTTERCUP_SCHEDULER_SERVE__COMPETITION_API_URL=http://competition-api:1323
       - BUTTERCUP_SCHEDULER_TASKS_STORAGE_DIR=/tasks_storage
       - BUTTERCUP_SCHEDULER_SCRATCH_DIR=/crs_scratch
       - BUTTERCUP_SCHEDULER_SERVE__MOCK_MODE=true
-      - BUTTERCUP_SCHEDULER_COMPETITION_API_URL=http://competition-api:1323
     depends_on:
       redis:
         condition: service_healthy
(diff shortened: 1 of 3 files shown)


orchestrator/src/buttercup/orchestrator/scheduler/vulnerabilities.py
```diff
@@ -71,13 +78,58 @@ def dedup_crash(self, crash: Crash) -> Crash | None:
...
         """
         logger.info(f"Submitting confirmed vulnerability for crash in {crash.target.package_name}")
-        # TODO: This is where we would submit the vulnerability to the competition api and get the vuln_id back
-        confirmed_vuln = ConfirmedVulnerability()
-        confirmed_vuln.crash.CopyFrom(crash)
-        confirmed_vuln.vuln_id = str(uuid.uuid4())  # TODO: ID got from the Competition API
-        self.confirmed_vulnerabilities_queue.push(confirmed_vuln)
+        try:
+            # Create vulnerability API client
+            vuln_api = VulnerabilityApi()
```
ret2libc: i think this can be created in the init/post-init. moreover, it should probably accept/require a user/password?

hbrodin: I agree.

compose.yaml
```diff
@@ -227,6 +228,8 @@ services:
       interval: 5s
       timeout: 5s
       retries: 10
+  competition-api:
```
ret2libc, line 234:
```suggestion
  
  competition-api:
```

orchestrator/src/buttercup/orchestrator/scheduler/config.py
```diff
@@ -43,6 +43,7 @@ class Settings(BaseSettings):
     serve: CliSubCommand[ServeCommand]
     process_ready_task: CliSubCommand[ProcessReadyTaskCommand]
     process_build_output: CliSubCommand[ProcessBuildOutputCommand]
+    competition_api_url: Annotated[str, Field(default="http://competition-api:8080", description="Competition API URL")]
```
ret2libc: should this be moved to `ServeCommand`?

PR #59, opened 2025-01-31
Challenge Task API

ret2libc:
This patch introduces a new `ChallengeTask` class (essentially a replacement of `OSSFuzzTool`) that can be used to interact with a challenges as stored by the CRS.

```python
task = ChallengeTask(task_dir, "example_project")
with task.get_rw_copy() as local_task:
   local_task.build_fuzzers(engine="libfuzzer", sanitizer="address")
   local_task.reproduce_pov("libpng_read_fuzzer", Path("/tmp/crash.bin"))
```

This is going to copy the task to a local temporary directory on the system, so the component can work on that instead of working on the shared folder which may be accessed by others.

For debugging/testing, there's also the option of working directly on a task without copying.

```python
task = ChallengeTask(task_dir, "example_project", local_task_dir=task_dir)
with task.get_rw_copy() as local_task:
   local_task.build_fuzzers(engine="libfuzzer", sanitizer="address")
   local_task.reproduce_pov("libpng_read_fuzzer", Path("/tmp/crash.bin"))
```

### DIFF
--- a/fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
+++ b/fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
@@ -38,10 +38,17 @@ def main():
             task_dir = Path(msg.source_path).parent.parent
             if args.allow_caching:
                 origin_task = ChallengeTask(
-                    task_dir, msg.package_name, python_path=args.python, local_task_dir=task_dir, logger=logger
+                    task_dir,
+                    msg.package_name,
+                    python_path=args.python,
+                    local_task_dir=task_dir,
                 )
             else:
-                origin_task = ChallengeTask(task_dir, msg.package_name, python_path=args.python, logger=logger)
+                origin_task = ChallengeTask(
+                    task_dir,
+                    msg.package_name,
+                    python_path=args.python,
+                )
 
             with origin_task.get_rw_copy(work_dir=args.wdir, delete=False) as task:
                 res = task.build_fuzzers_with_cache(engine=msg.engine, sanitizer=msg.sanitizer)


ret2libc:
I think this is ready for review! 

@2over12 could you have a look at the builder-bot part to see if you like how it looks with the new `ChallengeTask` class?

common/src/buttercup/common/challenge_task.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def __post_init__(self) -> None:
+        self.read_only_task_dir = Path(self.read_only_task_dir)
+        self.local_task_dir = Path(self.local_task_dir) if self.local_task_dir else None
+        self.python_path = Path(self.python_path)
+
+        if not self.read_only_task_dir.exists() or not self.read_only_task_dir.is_dir():
+            raise ValueError(f"Task directory does not exist: {self.read_only_task_dir}")
+
+        if self.local_task_dir is not None and (not self.local_task_dir.exists() or not self.local_task_dir.is_dir()):
```
hbrodin: Consider refactoring a lambda or something for
`not dir.exists() or not dir.is_dir()`

common/src/buttercup/common/challenge_task.py
```diff
@@ -0,0 +1,392 @@
...
+        # NOTE: assume the first directory inside the `src` subdir is the correct one
+        return self._find_first_dir(self.SRC_DIR)
+
+    def get_diff_subpath(self) -> Path | None:
+        # NOTE: assume the first directory inside the `diff` subdir is the correct one
+        return self._find_first_dir(self.DIFF_DIR)
+
+    def get_oss_fuzz_subpath(self) -> Path | None:
+        # NOTE: assume the first directory inside the `fuzz-tooling` subdir is the correct one
+        return self._find_first_dir(self.OSS_FUZZ_DIR)
```
hbrodin: Could we replace the NOTE with a reference to a github task were we track this?

common/src/buttercup/common/challenge_task.py
```diff
@@ -0,0 +1,392 @@
...
+
+@dataclass
+class CommandResult:
+    success: bool
+    error: bytes | None = None
+    output: bytes | None = None
+
+
+@dataclass
+class ChallengeTask:
```
hbrodin, line 30: This becomes quite a big class with somewhat diverse responsibilities. What are your thoughts on splitting it into two separate structures, one that deals with the paths for a challenge task and one that performs actions on a challenge task?

common/src/buttercup/common/utils.py
```diff
@@ -1,12 +1,45 @@
...
+import contextlib
+from contextlib import contextmanager
+from typing import Iterator, Any
+from tempfile import TemporaryDirectory
+from buttercup.common.logger import setup_logging
+from pathlib import Path
 
 
-def copyanything(src, dst):
+def copyanything(src: Path | str, dst: Path | str, **kwargs: Any) -> None:
```
hbrodin: `PathLike` ?

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
```diff
@@ -1,22 +1,19 @@
...
-import uuid
 import os
 
 logger = setup_logging(__name__)
 
 
 def main():
     prsr = argparse.ArgumentParser("Builder bot")
-    prsr.add_argument("--wdir", default=tempfile.TemporaryDirectory())
+    prsr.add_argument("--wdir", default="/tmp/builder-bot")
```
hbrodin, line 16: Why are we changing this? Should this be rooted in one of the existing dirs we have?

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
```diff
@@ -34,47 +31,39 @@ def main():
...
-                if not isinstance(wdirstr, str):
-                    wdirstr = args.wdir.name
-
-                target = os.path.join(wdirstr, f"ossfuzz-snapshot-{dirid}")
-                logger.info(f"Copying {ossfuzz_dir} to {target}")
-                shutil.copytree(ossfuzz_dir, target)
-                source_snapshot = os.path.join(wdirstr, f"source-snapshot-{dirid}-{msg.package_name}")
-                shutil.copytree(msg.source_path, source_snapshot)
-                source_path_output = source_snapshot
+            task_dir = os.path.dirname(os.path.dirname(msg.source_path))
```
hbrodin: Could this be simplified, or broken out as a named function to make it more clear what's going on?

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
```diff
@@ -34,47 +31,39 @@ def main():
...
+                )
+            else:
+                origin_task = ChallengeTask(task_dir, msg.package_name, python_path=args.python, logger=logger)
 
-            conf = BuildConfiguration(msg.package_name, msg.engine, msg.sanitizer, source_path_output)
-            logger.info(f"Building oss-fuzz project {msg.package_name}")
-            build_tool = OSSFuzzTool(Conf(target, args.python, args.allow_pull, args.base_image_url))
-            if not build_tool.build_fuzzer_with_cache(conf):
-                logger.error(f"Could not build fuzzer {msg.package_name}")
+            with origin_task.get_rw_copy(work_dir=args.wdir, delete=False) as task:
```
hbrodin, line 53: Say that we fail the build for whatever reason, an exception is thrown. Then this would be retried repeatedly, each time creating a copy of the full task dir.
Would it be an option to allow something like:
```
task.commit()
```
at the end when we know that we want to actually keep it?

Looking again, I assume that's basically what the `clean_on_failure` is for in `get_rw_copy`...

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
ret2libc: i'd like for the builder bot and the fuzzer to use `pathlib.Path` tbh as it makes things much easier to read, cc @2over12

ret2libc: it would also uniform with all the other components in the crs

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
ret2libc: yeah that's a good point... maybe a `task.commit()` approach would avoid having to clean things out with the `clean_on_failure` and maybe it would be clearer/more explicit when some piece of code wants to keep the new rw challenge task.

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
hbrodin: Make sense. I mostly reacted to the dirname(dirname(... thing

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
ret2libc: the reason for this was that tempfile.TemporaryDirectory() actually returns a context manager and not a temporary directory itself. (it should be used as `with TemporaryDirectory() as p: ...`).

moreover, this is just the base working directory, so other temporary dirs with random names are going to be created anyway, there's no need (I think) for the base dir itself to be also temporary and different every time.

common/src/buttercup/common/challenge_task.py
ret2libc: I'm not sure how to split those at this time. I would consider those as/if this class extends further, but for now a lot of the functions are related to performing actions on the challenge task anyway.

common/src/buttercup/common/challenge_task.py
```diff
@@ -0,0 +1,413 @@
...
+                pass
+
+    def commit(self, suffix: str | None = None, retry: bool = True) -> None:
+        """Commit the local task directory to a stable path.
+
+        This is useful to save the task state for later use and together with
+        the `get_rw_copy` context manager.
+        """
+        if self.local_task_dir is None:
+            return
```
hbrodin: I would consider raising an exception here. This is a programmer error, right?

common/src/buttercup/common/challenge_task.py
```diff
@@ -0,0 +1,413 @@
...
+            try:
+                self.logger.info(f"Committing task {self.local_task_dir} to {new_name}")
+                new_local_task_dir = self.local_task_dir.rename(self.local_task_dir.parent / new_name)
+                self.logger.info(f"Committed task {self.local_task_dir} to {new_name}")
+                break
+            except OSError:
+                if i == max_retries - 1:
+                    raise RuntimeError("Failed to commit task")
+
+                self.logger.error(
```
hbrodin: So, when would this happen? If we have hardcoded a suffix and that doesn't work I would almost expect us to raise an excpetion. I think that's a programmer error. If the suffix is random from the start then I guess there is some change it could collide but idk... What are your thoughts? I tend to favor hard stops when I consider it a programmer mistake.

common/src/buttercup/common/challenge_task.py
hbrodin: ok, let's leave it as is for now then.

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
hbrodin: Make sense.

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
hbrodin:
```suggestion
            task_dir = Path(msg.source_path).parent.parent
```
I think this would be more clear.

ret2libc:
- [x] use global logging instead of class one

PR #60, opened 2025-01-31
First cut fuzzer side crash dedup

2over12:
we should test this across a wide set of repros on the same harness ideally

### DIFF
--- a/fuzzer/src/buttercup/fuzzing_infra/fuzzer_bot.py
+++ b/fuzzer/src/buttercup/fuzzing_infra/fuzzer_bot.py
@@ -68,7 +68,7 @@ def main():
                     crash: engine.Crash = crash_
                     dst = crash_dir.copy_file(crash.input_path)
                     logger.info(f"Found crash {dst}")
-                    if not crash_set.add(
+                    if crash_set.add(
                         chc.target.package_name,
                         os.path.basename(chc.harness_path),
                         chc.target.task_id,


fuzzer/src/buttercup/fuzzing_infra/stack_parsing.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, redis: Redis):
+        self.redis = redis
+        self.set_name = "crash_set"
+        self.set = RedisSet(redis, self.set_name)
+
+    # Returns True if the crash was already in the set
+    def add(self, project: str, harness_name: str, task_id: str, stacktrace: str) -> bool:
+        crash_data = get_crash_data(stacktrace)
+        key = dumps([project, harness_name, task_id, crash_data], json_options=CANONICAL_JSON_OPTIONS)
+        return self.redis.sadd(self.set_name, key)
```
ret2libc:
```suggestion
        return self.set.add(self.set_name, key)
```

common/src/buttercup/common/sets.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class RedisSet:
+    def __init__(self, redis: Redis, set_name: str):
+        self.redis = redis
+        self.set_name = set_name
+
+    # Returns True if the value was already in the set
+    def add(self, value: str) -> bool:
+        return self.redis.sadd(self.set_name, value) == 0
```
ret2libc, line 12: this feels counterintuitive to me. i'd say it should return `true` when the element is really added, and false otherwise (it was already present).

common/src/buttercup/common/sets.py
```diff
@@ -0,0 +1,26 @@
...
+
+    # Returns True if the value was already in the set
+    def remove(self, value: str) -> bool:
+        return self.redis.srem(self.set_name, value) == 1
+
+    def contains(self, value: str) -> bool:
+        return self.redis.sismember(self.set_name, value)
+
+    def __iter__(self) -> Iterator[str]:
+        for member in self.redis.smembers(self.set_name):
```
ret2libc, line 22: is there the risk of loading everything into memory here?

fuzzer/src/buttercup/fuzzing_infra/fuzzer_bot.py
```diff
@@ -61,16 +62,25 @@ def main():
...
                 for crash_ in result.crashes:
                     crash: engine.Crash = crash_
                     dst = crash_dir.copy_file(crash.input_path)
                     logger.info(f"Found crash {dst}")
+                    if not crash_set.add(
+                        chc.target.package_name,
+                        os.path.basename(chc.harness_path),
+                        chc.target.task_id,
+                        crash.stacktrace,
+                    ):
```
ret2libc: ideally this could become just `crash_set.add(crash)`

common/src/buttercup/common/sets.py
```diff
@@ -0,0 +1,26 @@
...
+from redis import Redis
+
+
+class RedisSet:
+    def __init__(self, redis: Redis, set_name: str):
+        self.redis = redis
+        self.set_name = set_name
+
+    # Returns True if the value was already in the set
+    def add(self, value: str) -> bool:
```
ret2libc, line 11: a bit of llm-fu returned me an interesting approach for arbitrary canonical representation of Messages:
```
from google.protobuf.json_format import MessageToJson, Parse


    def _to_canonical_str(self, value: MsgType) -> str:
        # MessageToJson sorts fields by default and uses a deterministic representation
        json_str = MessageToJson(value, sort_keys=True, preserving_proto_field_name=True)
        # Ensure consistent whitespace by parsing and re-stringifying
        return json.dumps(json.loads(json_str), sort_keys=True)

    def _from_canonical_str(self, value_str: str) -> MsgType:
        msg = self.msg_builder()
        Parse(value_str, msg)
        return msg
```

there's also `including_default_value_fields` and `float_precision` that we could use for being even more precise with the JSON representation. The benefits of this would be we don't need to specify specific keys or use pymongo bson. Moreover, we can more easily create a `RedisSet(Generic[MsgType])` that we can reuse for any type of message. Moreover, same approach can be used for the FuzzerMap.

What do you think about this?

common/src/buttercup/common/sets.py
```diff
@@ -0,0 +1,26 @@
...
+        return self.redis.sadd(self.set_name, value) == 0
+
+    # Returns True if the value was already in the set
+    def remove(self, value: str) -> bool:
+        return self.redis.srem(self.set_name, value) == 1
```
ret2libc, line 16:
```suggestion
        return bool(self.redis.srem(self.set_name, value))
```

common/src/buttercup/common/sets.py
ret2libc:
```suggestion
    # Returns True if the value was already in the set
    def add(self, value: str) -> bool:
        return bool(self.redis.sadd(self.set_name, value))
```

common/src/buttercup/common/sets.py
2over12: yes

common/src/buttercup/common/sets.py
2over12: i dont think this is a good idea. there are multiple usecases where we do not want to include fields in the key. e.g. in the case of builds we want to dedup only on the fuzzer type to avoid allowing multiple engines/paths etc (that is what if we copy into two separate paths than dedup on the path, we end up with the same build saved)

2over12: what we dedup on is generally a design decision imo that needs to be thought about and we cant just do it for the complete message. I dont think any of the current maps or sets we have have the desired semantics of complete dedup

common/src/buttercup/common/sets.py
2over12: i think the semantics of adds contains makes sense. Im not really wiling to update a personal preference like this

common/src/buttercup/common/sets.py
ret2libc: i was not talking about dedup strategy, but just about creating a `RedisSet` data structure. that is, a data structure that has a `def add(o: objectClass)` method. Right now you are just manually storing strings, but this can be made more "protobuf-aware" with the methods above.

The dedup strategy is another story and should be solved elsewhere. In other words, suppose you have the regular `set()` from python... How you end up using that is different from how the set itself should be implemented.

PR #63, opened 2025-02-02
Allow multiple build types per harness

2over12:
This PR splits the recorded builds for a task and the harness map. This split allows separate information for a given harness, unifying corpuses across fuzzing engines, and multiple build types

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
+++ b/orchestrator/src/buttercup/orchestrator/scheduler/scheduler.py
@@ -16,6 +16,7 @@ from buttercup.common.datastructures.msg_pb2 import (
 from buttercup.orchestrator.scheduler.cancellation import Cancellation
 from buttercup.orchestrator.scheduler.vulnerabilities import Vulnerabilities
 from clusterfuzz.fuzz import get_fuzz_targets
+
 logger = logging.getLogger(__name__)
 
 


common/protos/msg.proto
```diff
@@ -61,17 +62,21 @@ message BuildOutput {
...
+    string build_type = 7;
 }
 
-message WeightedTarget {
+
+
+message WeightedHarness {
     float weight = 1;
-    BuildOutput target = 2; 
-    string harness_path = 3;
```
ret2libc, line 69: the path is always the same i guess, correct? e.g. build/out/<name>

common/src/buttercup/common/corpus.py
```diff
@@ -34,14 +34,12 @@ def copy_corpus(self, src_dir: str):
 
 
 class CrashDir(InputDir):
-    def __init__(self, harness_path: str):
-        build_dir = os.path.dirname(harness_path)
-        self.crash_dir = os.path.join(build_dir, f"{CRASH_DIR_NAME}_{os.path.basename(harness_path)}")
-        super().__init__(self.crash_dir)
+    def __init__(self, wdir: str, task_id: str, harness_name: str):
+        self.crash_dir = os.path.join(task_id, f"{CRASH_DIR_NAME}_{harness_name}")
```
ret2libc, line 38: general thing, but can we try to use `pathlib.Path` and related operations instead of various `os.path` methods? I think it's better to have a specific type for types and it's often easier to understand operations on the Path object rather than the os.path alternatives

common/src/buttercup/common/maps.py
```diff
@@ -31,23 +32,60 @@ def __iter__(self) -> Iterator[MsgType]:
...
 
-class FuzzerMap:
+class BUILD_TYPES(str, Enum):
+    FUZZER = "fuzzer"
+    COVERAGE = "coverage"
+
+
+# A build map makes it effecient to find for a given task_id + harness  a build type
+# we currently only support a single item of a given type
+# add a new type if you want to support dofferent builds
```
ret2libc:
```suggestion
# add a new type if you want to support different builds
```

common/src/buttercup/common/maps.py
```diff
@@ -31,23 +32,60 @@ def __iter__(self) -> Iterator[MsgType]:
...
+BUILD_MAP_NAME = "build_list"
 
 
-class FuzzerMap:
+class BUILD_TYPES(str, Enum):
+    FUZZER = "fuzzer"
+    COVERAGE = "coverage"
+
+
+# A build map makes it effecient to find for a given task_id + harness  a build type
```
ret2libc:
```suggestion
# A build map makes it effecient to find for a given task_id + harness a build type
```

common/src/buttercup/common/maps.py
```diff
@@ -31,23 +32,60 @@ def __iter__(self) -> Iterator[MsgType]:
...
+# we currently only support a single item of a given type
+# add a new type if you want to support dofferent builds
+class BuildMap:
+    def __init__(self, redis: Redis):
+        self.redis = redis
+
+    def map_key_from_task_id(self, task_id: str) -> str:
+        return dumps([task_id, BUILD_MAP_NAME], json_options=CANONICAL_JSON_OPTIONS)
+
```
ret2libc, line 53: not that it is a real problem, but is dumps+canonical etc. really needed here? i'd just do `return BUILD_MAP_NAME + "_" + task_id` or something like that.

common/src/buttercup/common/maps.py
```diff
@@ -31,23 +32,60 @@ def __iter__(self) -> Iterator[MsgType]:
...
-class FuzzerMap:
+class BUILD_TYPES(str, Enum):
+    FUZZER = "fuzzer"
+    COVERAGE = "coverage"
+
+
+# A build map makes it effecient to find for a given task_id + harness  a build type
+# we currently only support a single item of a given type
+# add a new type if you want to support dofferent builds
+class BuildMap:
```
ret2libc, line 47: can't we use a single `RedisMap` and use as key `task_id+build-type` and as value the buildoutput? AFAIU here you are using one map for each task and then the map will only have 2 items (either fuzzer or coverage).

ret2libc:
BTW shall FuzzerMap be moved to a `RedisSet` instead?

2over12:
> BTW shall FuzzerMap be moved to a `RedisSet` instead?

i dont think that works? we need to attach unkeyed data e.g. the weight and other data we want to find for a given harness

ret2libc:
> > BTW shall FuzzerMap be moved to a `RedisSet` instead?
> 
> i dont think that works? we need to attach unkeyed data e.g. the weight and other data we want to find for a given harness

what do you mean? those are part of the `WeightedTarget` object and right now the function is:
```python
    def push_target(self, target: WeightedTarget) -> None:
```

so it's essentially keeping track of all targets and we don't need access to a specific `WeightedTarget` by it's key (or do we?).

common/src/buttercup/common/maps.py
ret2libc: what about this?

ret2libc:
> > > BTW shall FuzzerMap be moved to a `RedisSet` instead?
> > 
> > 
> > i dont think that works? we need to attach unkeyed data e.g. the weight and other data we want to find for a given harness
> 
> what do you mean? those are part of the `WeightedTarget` object and right now the function is:
> 
> ```python
>     def push_target(self, target: WeightedTarget) -> None:
> ```
> 
> so it's essentially keeping track of all targets and we don't need access to a specific `WeightedTarget` by it's key (or do we?).

talked over slack, resolved!

common/src/buttercup/common/maps.py
2over12: i think we could the intention was to have a list of build types that we find given a task

2over12: i dont think it really matter tbh it may be nice to be able to iterate over all available build types for a given task. E.g. if you want to select all fuzzer engine types

PR #64, opened 2025-02-03
orchestrator/downloader: make task storage structure more rigid

ret2libc:
new task storage structure
```
<task-id>
<task-id>/src/
<task-id>/src/<dynamic-name-depending-on-src-tar-file>/...
<task-id>/fuzz-tooling/
<task-id>/fuzz-tooling/<dynamic-name-depending-on-fuzz-tar-file>/....
<task-id>/diff/
<task-id>/diff/<dynamic-name-depending-on-diff-tar-file>/....
```

In this way it's enough to point to a `<task-id>` director to know where to find all types of sources.

Moreover, it makes it easier to create a `ChallengeTask` api to build/reproduce/etc.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -166,7 +166,7 @@ services:
       - BUTTERCUP_SCHEDULER_LOG_LEVEL=debug
       - BUTTERCUP_SCHEDULER_SERVE__SLEEP_TIME=5
       - BUTTERCUP_SCHEDULER_TASKS_STORAGE_DIR=/tasks_storage
-      - BUTTERCUP_SCHEDULER_CRS_SCRATCH_DIR=/crs_scratch
+      - BUTTERCUP_SCHEDULER_SCRATCH_DIR=/crs_scratch
       - BUTTERCUP_SCHEDULER_SERVE__MOCK_MODE=true
     depends_on:
       redis:
--- a/orchestrator/test/test_scheduler.py
+++ b/orchestrator/test/test_scheduler.py
@@ -15,7 +15,7 @@ def mock_redis():
 
 @pytest.fixture
 def scheduler(mock_redis, tmp_path):
-    return Scheduler(tasks_storage_dir=tmp_path, crs_scratch_dir=tmp_path, redis=mock_redis)
+    return Scheduler(tasks_storage_dir=tmp_path, scratch_dir=tmp_path, redis=mock_redis)
 
 
 @pytest.mark.skip(reason="Not implemented")
--- a/orchestrator/src/buttercup/orchestrator/scheduler/__cli__.py
+++ b/orchestrator/src/buttercup/orchestrator/scheduler/__cli__.py
@@ -37,19 +37,19 @@ def main():
         redis = Redis.from_url(command.redis_url, decode_responses=False)
         scheduler = Scheduler(
             settings.tasks_storage_dir,
-            settings.crs_scratch_dir,
+            settings.scratch_dir,
             redis,
             sleep_time=command.sleep_time,
             mock_mode=command.mock_mode,
         )
         scheduler.serve()
     elif isinstance(command, ProcessReadyTaskCommand):
-        scheduler = Scheduler(settings.tasks_storage_dir, settings.crs_scratch_dir)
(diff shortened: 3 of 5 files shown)


orchestrator/src/buttercup/orchestrator/downloader/downloader.py
```diff
@@ -102,30 +113,24 @@ def safe_extract(tar, path: Path) -> None:
...
+                safe_extract(tar, destination)
 
                 # Extract all members directly into tmp_task_dir
                 for member in tar.getmembers():
-                    tar.extract(member, path=tmp_task_dir)
+                    tar.extract(member, path=destination)
 
             logger.info(f"[task {task_id}] Successfully extracted {source_file}")
-            # Remove the tar file after successful extraction
-            source_file.unlink()
```
hbrodin, line 116: keeping the tar file now?

orchestrator/test/test_scheduler.py
```diff
@@ -21,7 +21,7 @@ def scheduler(mock_redis, tmp_path):
 @pytest.mark.skip(reason="Not implemented")
 def test_process_ready_task(scheduler):
     # Create a mock task with example-libpng source
-    source = SourceDetail(source_type=SourceDetail.SourceType.SOURCE_TYPE_REPO, path="example-libpng")
+    source = SourceDetail(source_type=SourceDetail.SourceType.SOURCE_TYPE_REPO, url="example-libpng")
```
hbrodin: consider making actual urls

orchestrator/src/buttercup/orchestrator/downloader/downloader.py
ret2libc: nah, but it is in the temporary dir created in `_download_and_extract_sources`, thus it's going to be removed at that point.

2over12:
> Overall it looks good. However, aren't there any users oft he modified SourceDetail that need to be updated to locate the source? I would expect the fuzzer and builder to require changes.

i think we should be ok for now? the fuzzer currently relies on the source path information from the legacy build request message. once https://github.com/trailofbits/afc-crs-trail-of-bits/pull/59 is ready the legacy message will be replaced with just a task id and the challenge ID itself https://github.com/trailofbits/afc-crs-trail-of-bits/pull/59 can embed the heuristic for picking up source. This is i think we can wrap any changes into #59 once the ChallengeAPI is ready

ret2libc:
> once https://github.com/trailofbits/afc-crs-trail-of-bits/pull/59 is ready the legacy message will be replaced with just a task id

re what we discussed the other day, i think if we are not going to assume any "prefix path", we would have to pass around a path to the base task directory. Correct? @2over12

2over12:
> > once #59 is ready the legacy message will be replaced with just a task id
> 
> re what we discussed the other day, i think if we are not going to assume any "prefix path", we would have to pass around a path to the base task directory. Correct? @2over12

yeah maybe just a path to the task is best

compose.yaml
```diff
@@ -160,11 +160,13 @@ services:
...
     volumes:
       - ./crs_scratch:/crs_scratch
+      - ./tasks_storage:/tasks_storage
     environment:
       - BUTTERCUP_SCHEDULER_SERVE__REDIS_URL=redis://redis:6379
       - BUTTERCUP_SCHEDULER_LOG_LEVEL=debug
       - BUTTERCUP_SCHEDULER_SERVE__SLEEP_TIME=5
-      - BUTTERCUP_SCHEDULER_TASKS_STORAGE_DIR=/crs_scratch
+      - BUTTERCUP_SCHEDULER_TASKS_STORAGE_DIR=/tasks_storage
+      - BUTTERCUP_SCHEDULER_CRS_SCRATCH_DIR=/crs_scratch
```
hbrodin:
```suggestion
      - BUTTERCUP_SCHEDULER_SCRATCH_DIR=/crs_scratch
```

hbrodin: Nit: We aren't using the CRS-prefix elsewhere I think. I guess this require some additional updates, if accepted.

ret2libc: done. i might actually have to remove/rethink these vars to be honest, we might end up not using them as i was thinking about another approach but probably we are going to use just full paths everywhere.

PR #70, opened 2025-02-04
Adds a generic task loop structure for bots that want a harness and build type

2over12:


common/src/buttercup/common/default_task_loop.py
```diff
@@ -0,0 +1,55 @@
...
+            weighted_items: list[WeightedHarness] = self.harness_weights.list_harnesses()
+            logger.info(f"Received {len(weighted_items)} weighted targets")
+
+            if len(weighted_items) > 0:
+                chc = random.choices(
+                    weighted_items,
+                    weights=[it.weight for it in weighted_items],
+                    k=1,
+                )[0]
+                logger.info(f"Running fuzzer for {chc.harness_name} | {chc.package_name} | {chc.task_id}")
```
ret2libc: change this to something more generic

common/src/buttercup/common/default_task_loop.py
```diff
@@ -0,0 +1,55 @@
...
+
+                builds = dict(
+                    [(reqbuild, self.builds.get_build(chc.task_id, reqbuild)) for reqbuild in self.required_builds()]
+                )
```
ret2libc:
```suggestion
                builds = {reqbuild: self.builds.get_build(chc.task_id, reqbuild) for reqbuild in self.required_builds()}
```

common/src/buttercup/common/default_task_loop.py
```diff
@@ -0,0 +1,55 @@
...
+
+class TaskLoop(ABC):
+    def __init__(self, redis: Redis, timeout: int):
+        self.redis = redis
+        self.timeout = timeout
+        self.harness_weights = HarnessWeights(redis)
+        self.builds = BuildMap(redis)
+
+    # Declare a set of builds that must be available before running the task
+    def required_builds(self) -> List[BUILD_TYPES]:
```
ret2libc, line 22: shall we mark this as abstract as well?

2over12: it's not, it has a default implementation of no builds

ret2libc: yes that's what i mean. does it make sense to have no required builds? or to have a subclass that doesn't reimplement this method? it won't do anything in that case, so I think it should be mandatory to redefine this method.

PR #77, opened 2025-02-04
fuzzer: improve logging filtering

ret2libc:


2over12:
Is there a good reason for requiring passing down loggers to children instead of just using the global logger?

ret2libc:
> Is there a good reason for requiring passing down loggers to children instead of just using the global logger?

Because you can't know the user of the class what log level it wants. If you use the global logger the log level might be set before you actually set it from the CLI arguments.

ret2libc:
BTW if you use `@dataclass` things becomes even easier to understand and read/write.
See https://github.com/trailofbits/afc-crs-trail-of-bits/pull/59/files#diff-472393d52bbdc6f051c252c807042e2a6b8ca9dc36477a1fdc31985d21af1168R30 👍 

```python
   logger: logging.Logger = field(default_factory=lambda: setup_logging(__name__))
```

ret2libc:
Essentially if you use global logger, based on the order of the `import`, a children logger might get setup before the `fuzzer_bot` (or whatever other tool) has the chance to setup the correct logging level/handlers. If you know of any better way I'm open to suggestions. I might also ask on Slack.

ret2libc:
Ok talked with others, let's switch to global logging!

michaelbrownuc:
old, closing

PR #66, opened 2025-02-04
Send seed-gen seeds to fuzz corpus

reytchison:

PR #82, opened 2025-02-06
Add basic vuln discovery

reytchison:
Merge first: #66

### DIFF
--- a/seed-gen/src/buttercup/seed_gen/seed_gen_bot.py
+++ b/seed-gen/src/buttercup/seed_gen/seed_gen_bot.py
@@ -1,4 +1,5 @@
 import os
+import random
 import tempfile
 from pathlib import Path
 
@@ -9,7 +10,7 @@ from buttercup.common.datastructures.msg_pb2 import BuildOutput, WeightedHarness
 from buttercup.common.default_task_loop import TaskLoop
 from buttercup.common.logger import setup_logging
 from buttercup.common.maps import BUILD_TYPES
-from buttercup.seed_gen.tasks import do_seed_init
+from buttercup.seed_gen.tasks import Task, do_seed_init, do_vuln_discovery
 
 logger = setup_logging(__name__, os.getenv("LOG_LEVEL", "INFO").upper())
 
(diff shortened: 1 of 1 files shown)


seed-gen/src/buttercup/seed_gen/tasks.py
```diff
@@ -94,6 +63,17 @@ def do_seed_explore() -> None:
     raise NotImplementedError(f"{Task.SEED_EXPLORE} not implemented")
 
 
-def do_vuln_discovery() -> None:
+def do_vuln_discovery(challenge: str, output_dir: Path) -> None:
     """Do vuln-discovery task"""
-    raise NotImplementedError(f"{Task.VULN_DISCOVERY} not implemented")
+    logger.info("Doing vuln-discovery for challenge %s", challenge)
+    max_povs = 10
```
2over12, line 69: probably prefer to either make this a global constant or configured value

seed-gen/src/buttercup/seed_gen/utils.py
```diff
@@ -1,12 +1,44 @@
...
+logger = logging.getLogger(__name__)
+
 
 def resolve_module_subpath(subpath: str) -> Path:
     """Returns absolute path for file at subpath in module"""
     traversable = importlib.resources.files(f"buttercup.{__module_name__}").joinpath(subpath)
     return Path(str(traversable)).resolve()
+
+
+def extract_md(msg: AIMessage) -> str:
```
2over12, line 22: is there not a langchain output parser we can use?

seed-gen/src/buttercup/seed_gen/tasks.py
reytchison: Addressed in the migrated PR

seed-gen/src/buttercup/seed_gen/utils.py
reytchison: There is https://python.langchain.com/api_reference/langchain/output_parsers/langchain.output_parsers.regex.RegexParser.html which I looked at before. However I'm not sure if it's more convenient? I would like to stick with this helper function for now.

PR #83, opened 2025-02-06
Configure better logging system

ret2libc:


ret2libc:
@2over12 this is the best way to setup logging according to Will! WDYT?

common/src/buttercup/common/logger.py
```diff
@@ -1,12 +1,12 @@
...
 import os
 import tempfile
-from typing import Optional
 
 _is_initialized = False
+PACKAGE_LOGGER_NAME = "buttercup"
 
 
-def setup_logging(logger_name: Optional[str] = None, log_level: str = "info") -> logging.Logger:
+def setup_package_logger(logger_name: str, log_level: str = "info") -> logging.Logger:
```
hbrodin, line 9: Really nit: Is there a reason to not use the log_level constants? Also, I see we specify it as lower case string and do an upper later on. While still specifying the string "DEBUG" during invocation elsewhere.

ret2libc: it's to avoid doing the same operations on all CLI tools. This is almost always called by a CLI tool with an env var or with an argument, so it makes it easier to just call this function with that value directly without having to do `.upper()` or some mapping string->level in each tool.

PR #85, opened 2025-02-06
CI: add a workflow to run the whole CRS through compose

ret2libc:


ret2libc:
@hbrodin this is ready for review and it can actually be merged even if red, because it correctly runs everything!

.github/workflows/integration.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Build docker images
+        run: docker compose build
+
+      - name: Configure .env file
+        run: |
+          cp env.template .env
+          sed -i "s|AZURE_API_BASE=.*|AZURE_API_BASE=${{ secrets.AZURE_API_BASE }}|" .env
+          sed -i "s|AZURE_API_KEY=.*|AZURE_API_KEY=${{ secrets.AZURE_API_KEY }}|" .env
+
```
hbrodin, line 37: I didn't consider this before, but this will leave our credentials in the file, on disk. Now that we are using secrets, couldn't we just expose it as env-vars? e.g.
```
env:
    AZURE_API_BASE: ${{ secrets. AZURE_API_BASE }}
```
or is that somehow worse?

ret2libc: The compose file requires a .env file for `litellm`. In theory it can be changed, I guess, but... is it changing anything from a threat model point of view?

hbrodin: Idk. I think it is probably more "best practice" to not leave traces of the API-key on disk. But I guess it is not the end of the world. How hard would it be to just expose the env vars?
@2over12 any opinion?

PR #94, opened 2025-02-10
Henrik/patch submission

hbrodin:
This is an initial version of the path submission. It only implements the most basic patch submission functionality. There are no checks for duplicate submissions or other robustness featuers.


.github/workflows/integration.yml
```diff
@@ -53,6 +53,12 @@ jobs:
           while ! docker compose logs patcher | grep "Successfully generated patch for vulnerability libpng/"; do
             sleep 10
           done
+      - name: Wait for patch to be submitted
```
ret2libc, line 50: you can probably remove the previous step if this is now implement

PR #95, opened 2025-02-10
Port ASC patcher to AFC

ret2libc:
I've mostly taken the older patcher and ported to AFC, for now. That means there are a lot of things that need to be improved and fixed, but this works for now (with some mocking and special handling).

I plan to work on this more but this was the easiest way to have something almost fully functional immediately ready.

### DIFF
--- a/patcher/tests/test_patcher.py
+++ b/patcher/tests/test_patcher.py
@@ -35,8 +35,13 @@ def tasks_dir(tmp_path: Path) -> Path:
     yield tmp_path
 
 
-def test_vuln_to_patch_input(tasks_dir: Path):
-    patcher = Patcher(task_storage_dir=tasks_dir, redis=None, mock_mode=True)
+def test_vuln_to_patch_input(tasks_dir: Path, tmp_path: Path):
+    patcher = Patcher(
+        task_storage_dir=tasks_dir,
+        scratch_dir=tmp_path,
+        redis=None,
+        mock_mode=True,
+    )
 
     vuln = ConfirmedVulnerability(
         vuln_id="test-vuln-1",
@@ -64,7 +69,7 @@ def test_vuln_to_patch_input(tasks_dir: Path):
     assert patch_input.vulnerability_id == "test-vuln-1"
     assert patch_input.project_name == "libpng"
     assert patch_input.harness_name == "test-harness-name-1"
-    assert patch_input.pov == "test-crash-input-path-1"
+    assert patch_input.pov == Path("test-crash-input-path-1")
     assert patch_input.sanitizer_output == "test-stacktrace-1"
     assert patch_input.engine == "test-engine-1"
     assert patch_input.sanitizer == "test-sanitizer-1"


patcher/src/buttercup/patcher/utils.py
```diff
@@ -0,0 +1,61 @@
...
+from langchain_core.runnables import Runnable, RunnableConfig
+from typing import Callable
+
+VALID_PATCH_EXTENSIONS = (".c", ".h", ".in", ".java")
+
+CHAIN_CALL_TYPE = Callable[[Runnable, Callable, dict[str, Any], RunnableConfig | None, Any], Any]
+
+
+@dataclass
+class PatchInput:
```
hbrodin, line 20: Is this what should be received from the orchestrator, the part that is now a `ConfirmedVulnerability`?

patcher/src/buttercup/patcher/config.py
```diff
@@ -11,22 +11,22 @@ class ServeCommand(BaseModel):
 
 
 class ProcessCommand(BaseModel):
+    challenge_task_dir: CliPositionalArg[Path] = Field(description="Challenge Task Directory")
     task_id: CliPositionalArg[str] = Field(description="Task ID")
     vulnerability_id: CliPositionalArg[str] = Field(description="Vulnerability ID")
-    package_name: CliPositionalArg[str] = Field(description="Package Name")
+    project_name: CliPositionalArg[str] = Field(description="Package Name")
```
hbrodin, line 17: We should probably standardize on if we are using package or project or something else.

patcher/src/buttercup/patcher/agents/qe.py
```diff
@@ -0,0 +1,297 @@
...
+                        self.challenge.get_source_path().parent,
+                        "-i",
+                        patch_file.name,
+                        "--force",
+                        "--quiet",
+                        "--no-backup-if-mismatch",
+                    ],
+                    check=False,
+                )
+
```
hbrodin, line 214: Consider refactor to a function called twice with two different paths

patcher/src/buttercup/patcher/agents/rootcause.py
```diff
@@ -0,0 +1,518 @@
...
+        ("user", "Patch:\n```\n{patch}\n```"),
+        ("user", "Build failure stdout:\n```\n{build_stdout}\n```"),
+        ("user", "Build failure stderr:\n```\n{build_stderr}\n```"),
+    ]
+)
+
+
+@dataclass
+class RootCauseAgent:
+    """Software Engineer LLM agent, handling the creation of patches."""
```
hbrodin: Nit: Is this accurate? Seems like it doesn't align with the impl.

ret2libc: good catch, this is responsible for "triaging the vuln" essentially.

patcher/src/buttercup/patcher/utils.py
ret2libc: as discussed, the interface of the patcher does not need to change. This is just internal class because langchain/langfuse don't like dealing with protobuf classes directly.

PR #97, opened 2025-02-10
Update fuzzer to tasks 

2over12:


fuzzer/src/buttercup/fuzzing_infra/fuzzer_bot.py
```diff
@@ -35,48 +36,52 @@ def run_task(self, task: WeightedHarness, builds: dict[BUILD_TYPES, BuildOutput]
...
-            corp = Corpus(self.wdir, task.task_id, task.harness_name)
+            tsk = ChallengeTask(
+                read_only_task_dir=build.task_dir, project_name=build.package_name, python_path=self.python
+            )
 
-            copied_build_dir = os.path.join(td, os.path.basename(build_dir))
-            copied_corp_dir = os.path.join(copied_build_dir, corp.basename())
-            utils.copyanything(build_dir, copied_build_dir)
-            utils.copyanything(corp.path, copied_corp_dir)
+            with tsk.get_rw_copy(work_dir=td) as local_tsk:
```
ret2libc, line 44: do you still need the `with tempfile.TemporaryDirectory(dir=self.wdir) as td:` a few lines above? can't you create the corpus dir inside the `local_tsk`?

ret2libc: also, if we store the corpus inside the `local_tsk` itself, you could use the `build.task_dir` as the "main" corpus directory and when you do `get_rw_copy` it will automatically setup the local corpus (so no need for the extra copied_corp_dir later.

patcher/src/buttercup/patcher/patcher.py
```diff
@@ -108,7 +108,7 @@ def process_vulnerability(self, input: PatchInput) -> PatchOutput | None:
     def _create_patch_input(self, vuln: ConfirmedVulnerability) -> PatchInput:
         return PatchInput(
             # TODO: modify to use proper task_dir once in msg
```
ret2libc: please remove this TODO as well now

fuzzer/src/buttercup/fuzzing_infra/fuzzer_bot.py
2over12: the corpus thing does not work because corpuses are shared across multiple builds

PR #128, opened 2025-07-03
Allow for non AIxCC-related deployment

ret2libc:
A few things I had to do to make this work:
- push https://github.com/orgs/trailofbits/packages/container/package/buttercup-cscope 
- copy the competition-test-api from aixcc-finals into https://github.com/orgs/trailofbits/packages/container/package/buttercup%2Fcompetition-test-api

Unfortunately, although the competition-test-api works for submitting tasks to the CRS, the whole workflow won't work well because it is going to reject any PoV when found by the CRS (it expects some aixcc-specific flags in oss-fuzz that are ofc not present in the upstream version).

### DIFF
--- a/deployment/crs-architecture.sh
+++ b/deployment/crs-architecture.sh
@@ -64,6 +64,15 @@ up() {
 	export TS_DNS_IP
 
 	if [ "$DEPLOY_CLUSTER" = "true" ]; then
+		# Authenticate with GitHub Container Registry for Docker builds
+		if [ -n "$GHCR_AUTH" ]; then
+			echo -e "${BLU}Authenticating with GitHub Container Registry${NC}"
+			GHCR_PAT=$(echo -n "$GHCR_AUTH" | base64 -d | cut -d: -f2)
+			GHCR_USERNAME=$(echo -n "$GHCR_AUTH" | base64 -d | cut -d: -f1)
+		else
+			echo -e "${RED}Warning: GHCR_AUTH not set, Docker builds may fail to pull from ghcr.io${NC}"
+		fi
+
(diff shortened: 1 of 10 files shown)


ret2libc:
I've created a 1p vault called `buttercup`, where I've uploaded some new secrets completely separated from what we have used for aixcc and a env file that can be used *almost* as a drop-in. The only thing that needs to be adjusted is the `GHCR_AUTH`, because I don't want to put my personal token there.

hbrodin:
We still rely on aixcc images. I was under the impression that we would replace those with upstream images. Is this an intermediate step or am I missing something?

ret2libc:
> We still rely on aixcc images. I was under the impression that we would replace those with upstream images. Is this an intermediate step or am I missing something?

Sorry maybe it's not clear as I couldn't share the env file, but in the env file you can now specify the base image of a few things, see https://github.com/trailofbits/buttercup/pull/128/files#diff-54fe07d4aa89ef451e05e8f4961416f485e9ad6db0d8c2c84dfec4e5e2de305dR112-R126 .

hbrodin:
oh, ok. That makes sense. I think I was expecting those to be the new defaults.

ret2libc:
> oh, ok. That makes sense. I think I was expecting those to be the new defaults.

if we want to completely move away from aixcc, i can do that yes!

hbrodin:
> > oh, ok. That makes sense. I think I was expecting those to be the new defaults.
> 
> if we want to completely move away from aixcc, i can do that yes!

I think either way is fine. I guess eventually we want to make it standalone, don't see much point in keeping the aixcc infra. However, there is nothing wrong with doing so incrementally.

PR #129, opened 2025-07-03
common/util-cli: provide some stats in read_submissions

ret2libc:
These are just some small improvements to the utility script we were using to monitor internal dry runs and see how we were doing. The PR adds some stats.

### DIFF
--- a/common/src/buttercup/common/util_cli.py
+++ b/common/src/buttercup/common/util_cli.py
@@ -33,6 +33,7 @@ class TaskResult(BaseModel):
     mode: str
     n_vulnerabilities: int = 0
     n_patches: int = 0
+    n_bundles: int = 0
     patched_vulnerabilities: list[str] = []
     non_patched_vulnerabilities: list[str] = []
 
@@ -242,6 +243,10 @@ def main():
                     if c:
                         result[task_id].non_patched_vulnerabilities.append(c.competition_pov_id)
 
+                b = next((b for b in submission.bundles), None)
+                if b:
+                    result[task_id].n_bundles += 1
+
                 print(f"--- Submission {i} ---")
                 print(submission)
                 print()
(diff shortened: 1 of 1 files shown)


common/src/buttercup/common/util_cli.py
```diff
@@ -200,12 +220,75 @@ def main():
...
+                    logger.error(f"Task {task_id} not found in registry")
+                    continue
+
+                if task_id not in result:
+                    result[task_id] = TaskResult(
+                        task_id=task_id, project_name=task.project_name, mode=str(task.task_type)
+                    )
+                c = next((c for c in submission.crashes if c.result == SubmissionResult.PASSED), None)
+                if c:
+                    result[task_id].n_vulnerabilities += 1
```
hbrodin, line 236: This could only ever become one I think? In theory we could have multiple `PASSED` vulns for a single `SubmissionEntry`.

common/src/buttercup/common/util_cli.py
```diff
@@ -200,12 +220,75 @@ def main():
...
+
+                p = next((p for p in submission.patches if p.result == SubmissionResult.PASSED), None)
+                if p:
+                    result[task_id].n_patches += 1
```
hbrodin, line 240: Same here.

common/src/buttercup/common/util_cli.py
```diff
@@ -200,12 +220,75 @@ def main():
...
+                    result[task_id].patched_vulnerabilities.append(c.competition_pov_id)
```
hbrodin, line 241: The connection between patch and PoV is primarily maintained in the bundle connected to the SubmissionEntry. If we have a PoV and Patch both being ACCEPTED, we should see a bundle real soon. Consider getting the information from there. Also, there might be multiple PoVs and Patches in the PASSED state for a single SubmissionEntry (as noted above) due to merging. While those should all represent the same thing it might be more accurate to append the full list of vulns and patches if they are in the passed state.

common/src/buttercup/common/util_cli.py
ret2libc: Why? There can be multiple vulnerabilities in the same task, so this could be higher.

common/src/buttercup/common/util_cli.py
ret2libc: Same answer. There could be multiple vulns for a task, thus multiple patches

common/src/buttercup/common/util_cli.py
hbrodin: ok, sorry, bad phrasing. For each `SubmissionEntry` there could in theory be more than one vuln that is in the `PASSED` state., but using `next` will just give us one of these. We would need something like `sum(1 for c in submissions.crashes if ...`) and `if c > 0: n_vulnerabilities += c` or something to be more accurate.

PR #137, opened 2025-07-07
deployment: fix CRS/Service instance IDs usage

ret2libc:


ret2libc:
We were using SERVICE_INSTANCE_ID wrong, so in the last week i fixed it, but I didn't want to make too many changes, so I just changed crs <-> service -instance-id in the minimum places possible. Now I'm completing that change so that things are clearer.

PR #139, opened 2025-07-07
common: fix mypy

ret2libc:


common/src/buttercup/common/node_local.py
ret2libc, line 1: @hbrodin could you have a look at this?

common/src/buttercup/common/challenge_task.py
```diff
@@ -171,12 +176,16 @@ def __post_init__(self) -> None:
                 raise ChallengeTaskError(f"Missing required directory: {self.task_dir / directory}")
 
         self._helper_path = Path("infra/helper.py")
-        if not (self.get_oss_fuzz_path() / self._helper_path).exists():
-            raise ChallengeTaskError(f"Missing required file: {self.get_oss_fuzz_path() / self._helper_path}")
+        oss_fuzz_path = self.get_oss_fuzz_path()
```
hbrodin, line 182: Another option would be to raise in `get_oss_fuzz_path` instead. That way we either have a value or an exception and we don't need all of these `if oss_fuzz_path is None:`

common/src/buttercup/common/node_local.py
```diff
@@ -89,22 +89,28 @@ def scratch_path() -> NodeLocalPath:
...
-def scratch_dir() -> TmpDir:
+@contextmanager
+def scratch_dir() -> Iterator[TmpDir]:
     """Return a temporary directory in the scratch directory"""
-    return temp_dir(scratch_path())
+    with temp_dir(scratch_path()) as tmp_dir:
+        yield tmp_dir
 
 
-def local_scratch_file(**kwargs) -> NamedTemporaryFile:
```
hbrodin, line 97: Is there actually a problem with this code? The change looks good, but I believe the existing behavior is still valid. Do you agree?

common/src/buttercup/common/node_local.py
```diff
@@ -89,22 +89,28 @@ def scratch_path() -> NodeLocalPath:
...
+@contextmanager
+def local_scratch_file(**kwargs: Any) -> Iterator[Any]:
     """Return a temporary file in the local scratch directory"""
     sp = scratch_path()
-    return NamedTemporaryFile(dir=sp, **kwargs)
+    with NamedTemporaryFile(dir=sp, **kwargs) as tmp_file:
+        yield tmp_file
 
 
-def remote_scratch_file(local_path: NodeLocalPath, **kwargs) -> NamedTemporaryFile:
```
hbrodin, line 103: Same here.

common/src/buttercup/common/queues.py
```diff
@@ -203,16 +134,24 @@ def push(self, item: MsgType) -> None:
         bts = item.SerializeToString()
         self.redis.xadd(self.queue_name, {self.INAME: bts})
 
-    def _ensure_group_name(func):
-        def wrapper(self, *args, **kwargs):
+    @staticmethod
+    def _ensure_group_name[**P, R](
```
hbrodin: Good to have explicit types but this starts to look like Rust now :)

ret2libc: :P

ret2libc:
(there are some extra changes included, we should merge #147 first)

michaelbrownuc:
I merged #151 recently which has caused a number of dependency mismatches in this PR -  @ret2libc sorry but can you double check these?  If it were a few I'd fix them them myself but there are many changes and I'm not sure if any will break this PR.

ret2libc:
@michaelbrownuc updated

PR #141, opened 2025-07-08
buttercup-ui: basic interface to the CRS

ret2libc:


ret2libc:
With this we can already replace the existing competition-api and submit custom challenges through the regular interface (see `scripts/task_crs.sh` for example). When submitting PoVs/patches/bundles it will just answer with a positive answer, so that the CRS can move forward.

deployment/k8s/templates/config.yaml
```diff
@@ -15,5 +15,5 @@ metadata:
 data:
   CRS_API_KEY_ID: "{{ .Values.global.crs.api_key_id }}"
   CRS_API_URL: "{{ .Values.global.crs.api_url | default (printf "http://%s-task-server.%s.svc.cluster.local:8000" .Release.Name .Release.Namespace) }}"
-  COMPETITION_API_URL: "{{ .Values.global.crs.competition_api_url | default (printf "http://%s-competition-api.%s.svc.cluster.local:1323" .Release.Name .Release.Namespace) }}"
+  COMPETITION_API_URL: "{{- $raw := .Values.global.crs.competition_api_url -}}{{- if $raw -}}{{- if hasPrefix "http" $raw -}}{{ $raw }}{{- else -}}{{ printf "http://%s-%s.%s.svc.cluster.local:1323" .Release.Name $raw .Release.Namespace }}{{- end -}}{{- else -}}{{ printf "http://%s-competition-api.%s.svc.cluster.local:1323" .Release.Name .Release.Namespace }}{{- end -}}"
```
evandowning: Do we want to be more precise by using `if hasPrefix "http://"` instead?

orchestrator/src/buttercup/orchestrator/ui/README.md
```diff
@@ -0,0 +1,97 @@
...
+│   ├── models/              # Pydantic models for API and CRS
+│   └── services/            # Business logic (challenge, CRS, etc.)
+└── ...
+```
+
+## Configuration
+
+The UI is configured via environment variables (with the `BUTTERCUP_UI_` prefix) or a `.env` file. Key settings include:
+
+- `BUTTERCUP_UI_HOST`: Host to bind the server (default: `0.0.0.0`)
```
evandowning: Doesn't this expose buttercup to the wider internet? Should we constrain it to default to `127.0.0.1` instead?

orchestrator/src/buttercup/orchestrator/ui/config.py
```diff
@@ -0,0 +1,42 @@
...
+    # Server configuration
+    redis_url: Annotated[str, Field(default="redis://localhost:6379", description="Redis URL")]
+    log_level: Annotated[str, Field(default="debug", description="Log level")]
+    reload: Annotated[bool, Field(default=False, description="Reload the server when code changes")]
+
+    # Competition API configuration
+    external_host: Annotated[
+        str, Field(default="localhost", description="External host, used to construct the files URLs")
+    ]
+    host: Annotated[str, Field(default="0.0.0.0", description="Host to bind the server to")]
```
evandowning: Similar comment

PR #142, opened 2025-07-08
feat: implement REST API for program-model component

### LINKED ISSUE #116: Create REST api interface to program-model
Instead of executing cqsearch/codequery from seed-gen/patcher, we should use a REST API so we have better isolation.

evandowning:
This PR implements a comprehensive REST API for the program-model component to improve isolation between services.

## Changes
- Add FastAPI server with endpoints for all CodeQuery operations
- Create HTTP client and compatibility layer for seamless migration
- Update patcher and seed-gen to use REST API instead of direct execution
- Remove codequery/cscope dependencies from Dockerfiles
- Move harness discovery functions to program-model component
- Add comprehensive unit tests and API documentation
- Provide CLI script for starting the API server

Resolves #116

-----------------------------------

**TODOs**:

- [x] Run `make test`
- [x] Run libpng challenge
- [ ] Confirm modified `patcher` and `seed-gen` unit tests make sense.
- [x] Merge #148 into `main` and rebase

### DIFF
--- a/program-model/src/buttercup/program_model/api/server.py
+++ b/program-model/src/buttercup/program_model/api/server.py
@@ -26,6 +26,9 @@ from buttercup.program_model.api.models import (
     ErrorResponse,
     FunctionModel,
     FunctionSearchResponse,
+    HarnessInfoModel,
+    HarnessSearchRequest,
+    HarnessSearchResponse,
     TaskInitRequest,
     TaskInitResponse,
     TypeDefinitionModel,
@@ -33,6 +36,12 @@ from buttercup.program_model.api.models import (
     TypeUsageInfoModel,
 )
 from buttercup.program_model.codequery import CodeQueryPersistent
+from buttercup.program_model.harness_finder import (
+    find_libfuzzer_harnesses,
+    find_jazzer_harnesses,
+    get_harness_source_candidates,
+    get_harness_source,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -181,6 +190,14 @@ async def initialize_task(
 ) -> TaskInitResponse:
     """Initialize a CodeQuery instance for a task."""
     try:
+        if task_id in _codequery_instances:
+            logger.info("Task %s already initialized", task_id)
+            return TaskInitResponse(
+                task_id=task_id,
+                status="already_initialized",
+                message="Task already initialized",
+            )
+
         task_dir = Path(request.task_dir)
         work_dir = Path(request.work_dir)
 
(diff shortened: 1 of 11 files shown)


seed-gen/src/buttercup/seed_gen/task.py
```diff
@@ -121,7 +121,13 @@ def get_llm(llm: ButtercupLLM, fallback_llms: list[ButtercupLLM]) -> BaseChatMod
         return llm.with_fallbacks(fallbacks)
 
     def get_harness_source(self) -> HarnessInfo | None:
```
reytchison, line 123: seed-gen currently fails to find harnesses:
```
2025-07-22 17:59:05,389 - buttercup.common.utils - INFO - Getting diffs from /node_data/crs_scratch/ebd1e20c-ebe2-41c1-985d-3b2162a582d4/seedgen-jrt7y991/ebd1e20c-ebe2-41c1-985d-3b2162a582d4/ebd1e20c-ebe2-41c1-985d-3b2162a582
d4-3318dfe0-f58b-4e-7jsg71jl/diff/diff                                                                                                                                                                                           
2025-07-22 17:59:05,389 - buttercup.seed_gen.vuln_base_task - INFO - Doing vuln-discovery for challenge libpng (mode: delta)                                                                                                     
2025-07-22 17:59:05,395 - buttercup.seed_gen.vuln_base_task - ERROR - Failed vuln-discovery for challenge libpng: ('No harness found for challenge %s', 'libpng')                                                                
Traceback (most recent call last):                                                                                                                                                                                               
  File "/app/seed-gen/.venv/lib/python3.12/site-packages/buttercup/seed_gen/vuln_base_task.py", line 323, in do_task                                                                                                             
    state = self._init_state(out_dir, current_dir)                                                                                                                                                                               
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                               
  File "/app/seed-gen/.venv/lib/python3.12/site-packages/buttercup/seed_gen/vuln_discovery_delta.py", line 96, in _init_state                                                                                                    
    raise ValueError("No harness found for challenge %s", self.package_name)                                                                                                                                                     
ValueError: ('No harness found for challenge %s', 'libpng')   
```

reytchison: It appears that all seed-gen tasks are failing because of this

reytchison: I'm not sure it is worth making harness lookup over the REST API. It is program contextualization but seed-gen is the only component that uses it. Also, the current PR doesn't free seed-gen from the 1 harness lookup dependency (ripgrep) because it allows seed-gen to lookup harnesses without the API.

program-model/src/buttercup/program_model/harness_models.py
```diff
@@ -0,0 +1,23 @@
+"""Models for harness-related functionality."""
+
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass
+class HarnessInfo:
```
reytchison, line 8: This should be internal to seed-gen because it's seed-gen prompting. (see the __str__ method).

seed-gen/src/buttercup/seed_gen/task.py
```diff
@@ -21,7 +21,7 @@
 from buttercup.common.challenge_task import ChallengeTask
 from buttercup.common.llm import ButtercupLLM, create_default_llm, get_langfuse_callbacks
 from buttercup.common.project_yaml import ProjectYaml
-from buttercup.program_model.codequery import CodeQueryPersistent
+from buttercup.program_model.rest_client import CodeQueryPersistentRest as CodeQueryPersistent
```
reytchison, line 24: Nit: import as CodeQueryPersistentRest instead of renaming to CodeQueryPersistent

seed-gen/src/buttercup/seed_gen/seed_gen_bot.py
```diff
@@ -136,7 +136,7 @@ def run_task(self, task: WeightedHarness, builds: dict[BuildTypeHint, list[Build
 
             logger.info("Initializing codequery")
             try:
-                codequery = CodeQueryPersistent(challenge_task, work_dir=Path(self.wdir))
+                codequery = CodeQueryPersistentRest(challenge_task, work_dir=Path(self.wdir))
```
reytchison, line 139: I don't think it makes sense for the client to specify the working directory for the server? The work_dir is passed to the server via a `TaskInitRequest`, which the server then opens. Shouldn't this be transparent to the client? It also lets the client overwrite arbitrary directories in the server.

program-model/src/buttercup/program_model/api/server.py
```diff
@@ -0,0 +1,678 @@
...
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],  # Configure as needed for production
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
+
+# Global storage for CodeQuery instances
+_codequery_instances: Dict[str, CodeQueryPersistent] = {}
```
reytchison, line 64: I don't think uvicorn shares memory across workers so if there are multiple workers (which can be configured in the CLI) then this will be per worker and each worker will have to maintain its own codequerypersistent instance.

reytchison: Also this means if there are multiple workers then the behavior of the search methods will depend on whether that worker has initialized a CodeQueryPersistent for that task. If it hasn't, then the search will fail.

program-model/src/buttercup/program_model/api/server.py
```diff
@@ -0,0 +1,678 @@
...
+
+
+@app.post("/tasks/{task_id}/init", response_model=TaskInitResponse)
+async def initialize_task(
+    request: TaskInitRequest,
+    task_id: str = FastAPIPath(..., description="Task ID"),
+) -> TaskInitResponse:
+    """Initialize a CodeQuery instance for a task."""
+    try:
+        if task_id in _codequery_instances:
```
reytchison, line 193: Ignores the workdir and challenge_task dir in the request if a CodeQueryPersistent instance already exists for the task ID. I think this another reason for just passing task_id instead of workdir and challenge_task dir.

program-model/src/buttercup/program_model/api/server.py
```diff
@@ -0,0 +1,678 @@
...
+            logger.exception("Failed to initialize ChallengeTask: %s", e)
+            raise HTTPException(
+                status_code=500,
+                detail=f"Failed to initialize ChallengeTask: {str(e)}",
+            )
+
+        # Initialize CodeQuery
+        logger.info("Creating CodeQueryPersistent...")
+        try:
+            codequery = CodeQueryPersistent(
```
reytchison, line 255: If a codequery db doesn't exist yet then I think the API server will create the codequery DB, similar to what seed-gen and the patcher did before. If the server is a single worker then this will block lookups for other task IDs, which may be initialized.

program-model/src/buttercup/program_model/api/server.py
reytchison: Also this will not be shared across the program model replicas (there are 4 in the prod configuration)

deployment/k8s/charts/program-model/templates/api-deployment.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if .Values.api.enabled }}
```
reytchison, line 1: Might need to scale the API for a production deployment? The production deployment has 4 program model replicas and each replica has 1 API pod with 1 uvicorn worker. However the prod deployment has 20 seed-gen workers and 28 patchers. It seems like they could saturate the API workers

program-model/src/buttercup/program_model/api/server.py
```diff
@@ -0,0 +1,678 @@
...
+    if missing_commands:
+        logger.error("Missing required commands: %s", ", ".join(missing_commands))
+        raise RuntimeError(f"Missing required commands: {', '.join(missing_commands)}")
+
+    # Check environment
+    logger.info("Current working directory: %s", Path.cwd())
+    logger.info("Docker socket available: %s", Path("/var/run/docker.sock").exists())
+
+    # Check if we can access common directories
+    common_dirs = ["/crs_scratch", "/tasks_storage", "/node_data"]
```
reytchison, line 116: I'm not sure if this check is necessary and these paths should be hardcoded?

program-model/src/buttercup/program_model/api/server.py
```diff
@@ -0,0 +1,678 @@
...
+
+
+async def cleanup_file(filepath: Path) -> None:
+    """A function to clean up a file."""
+    if filepath.exists():
+        filepath.unlink()
+
+
+@app.get("/tasks/{task_id}/container-src-dir")
+async def download_container_src_dir(
```
reytchison, line 633: This endpoint is unused in the client/elsewhere.

Issue #127, opened 2025-07-03
Modified cscope dependency


ret2libc:
Right now the CRS depends on a slightly modified version of cscope (https://github.com/trailofbits/aixcc-cscope), which was pushed on the aixcc-finals repository as a package so it could be accessed by other containers.

This is a dependency that would be good to remove to make the whole repository settings and deployment easier, without extra deps.

I've already pushed the container under our own organization, so we can use that package/container instead of the one hosted in aixcc-finals repository, but still.

PR #148, opened 2025-07-09
Build cscope image locally instead of pulling

reytchison:
Add aixcc-cscope as a submodule and build it locally instead of pulling

Also since aixcc-cscope is private, we need to use a token to checkout the project in CI. The secret is currently a fine-grained PAT. We can't use the default GITHUB_TOKEN which is passed to the checkout action: https://github.com/actions/checkout?tab=readme-ov-file#checkout-multiple-repos-private

Closes: https://github.com/trailofbits/buttercup/issues/127

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "external/aixcc-cscope"]
 	path = external/aixcc-cscope
-	url = git@github.com:trailofbits/aixcc-cscope.git
+	url = https://github.com/trailofbits/aixcc-cscope.git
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = git@github.com:trailofbits/aixcc-kythe.git
 [submodule "external/aixcc-cscope"]
 	path = external/aixcc-cscope
-	url = https://github.com/trailofbits/aixcc-cscope.git
+	url = git@github.com:trailofbits/aixcc-cscope.git
```
ret2libc: we need the `buttercup` branch of cscope though

reytchison: The submodule is checked out at that commit, see https://github.com/trailofbits/buttercup/pull/148/files#diff-e9f068cf1fa9fc3d72af0e7a31bd9588f196aa9d3ba05513981a1985d3c6452e

PR #152, opened 2025-07-10
common: add first frame line number in the crash set

### LINKED ISSUE #134: Look into solving integration-test challenge
We currently can't solve the integration-test challenge https://github.com/aixcc-finals/integration-test/tree/challenges/integration-test-delta-01 because of how the fuzzer/tracer bots work.

In a few words, the problem is that there are 2 vulns in the same function, one is introduced by the delta, the other is not. If the one *not* introduced by the delta is detected first, the tracer-bot is going to consider it invalid. At the same time, the crash-token is added to the CrashSet, thus all future crashes with the same crash-token (like the other vuln in the same function) are going to be directly discarded in the fuzzer and they will never be re-tested.

This leads to us missing a valid vulnerability.

ret2libc:
This will differentiate between vulnerabilities in different lines of the same function, so that they all go to the tracer-bot for validation. In this way, if there is a DELTA mode challenge and one vuln is present both before and after the delta, but another one is introduced by the delta, we are able to validate them separately and actually find the vulnerability that is introduced by the delta (and not discard it in the fuzzer-bot due to being already present in the CrashSet).

It means the tracer-bot will receive slightly more crashes to validate, but still a reasonable number.

Closes #134


ret2libc:
@michaelbrownuc what do you think about this? I'll leave this up to you, as it slightly changes the behaviour of the system. As said, I think it's not going to change much, as we would still consider these PoVs as similar in the orchestrator, because they have a have similar crash_states (and i'm not touching the crash states, just changing how `CrashSet` works. `CrashSet` is used only by seed-gen and fuzzer).

However, what this does is that we won't discard valid PoVs just because we found an invalid one before at a different line. I personally think it's worthwhile.

ret2libc:
cc @evandowning @michaelbrownuc if you choose to not merge this, consider changing the test challenge in https://github.com/trailofbits/buttercup/pull/153/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R95 .

Issue #165, opened 2025-07-23
`make local-setup` should not fail if run multiple times


BradSwain:
Running `make local-setup` should succeed if everything is already setup for local development.

If `make local-setup` is cancelled or fails part way through, re-running may fail, forcing users to fall back to manual setup.

For example, the local setup fails if just is already installed.

```bash
brad@computer:~/buttercup$ make setup-local
Setting up local development environment...
./scripts/setup-local.sh
🚀 Setting up Buttercup CRS for local development...
[INFO] Starting local development setup...
[INFO] Installing Docker...
[SUCCESS] Docker is already installed
[INFO] Installing kubectl...
    ...
[SUCCESS] Just installed successfully
[INFO] Setting up configuration...
[SUCCESS] Configuration file created from template
[INFO] Configuring required API keys for local development...
Enter your OpenAI API key: ^C Interrupt

brad@computer:~/buttercup$ make setup-local
Setting up local development environment...
./scripts/setup-local.sh
🚀 Setting up Buttercup CRS for local development...
[INFO] Starting local development setup...
[INFO] Installing Docker...
[SUCCESS] Docker is already installed
[INFO] Installing kubectl...
[SUCCESS] kubectl is already installed
[INFO] Installing Helm...
[SUCCESS] Helm is already installed
[INFO] Installing Minikube...
[SUCCESS] Minikube is already installed
[INFO] Installing Git LFS...
[SUCCESS] Git LFS is already installed
[INFO] Installing Just...
install: Repository:  https://github.com/casey/just
install: Crate:       just
install: Tag:         1.42.3
install: Target:      x86_64-unknown-linux-musl
install: Destination: /home/brad/bin
install: Archive:     https://github.com/casey/just/releases/download/1.42.3/just-1.42.3-x86_64-unknown-linux-musl.tar.gz
install: error: `/home/brad/bin/just` already exists
make: *** [Makefile:33: setup-local] Error 1
```

PR #168, opened 2025-07-23
allow updating values in `make setup-local`

BradSwain:
related to #165

Allows `make setup-local` to be run multiple times to update/unset values. Also adds a short description for each value.

For example:

Skipping a value looks like:
```bash
OpenAI API Key (Optional): Powers AI-driven vulnerability analysis and patch generation.
The patcher component performs best with OpenAI models (GPT-4o/GPT-4o-mini).
Enter your OpenAI API key (press Enter to skip): 
[INFO] OpenAI API key disabled
```

Setting a value looks like:
```bash
Anthropic API Key (Optional): Powers AI-driven fuzzing seed generation.
The seed generation component performs best with Anthropic models (Claude 3.5/4 Sonnet).
Enter your Anthropic API key (press Enter to skip): 
[SUCCESS] Anthropic API key configured
```

If `make setup-local` is run again the value can be kept:
```bash
OpenAI API Key (Optional): Powers AI-driven vulnerability analysis and patch generation.
The patcher component performs best with OpenAI models (GPT-4o/GPT-4o-mini).
OpenAI API key is already configured. Set a new value? (y/N): 
[INFO] Keeping existing OpenAI API key
```

updated:
```bash
Anthropic API Key (Optional): Powers AI-driven fuzzing seed generation.
The seed generation component performs best with Anthropic models (Claude 3.5/4 Sonnet).
Anthropic API key is already configured. Set a new value? (y/N): y
Enter your Anthropic API key (press Enter to skip): 
[SUCCESS] Anthropic API key configured
```

or cleared:
```bash
Anthropic API Key (Optional): Powers AI-driven fuzzing seed generation.
The seed generation component performs best with Anthropic models (Claude 3.5/4 Sonnet).
Anthropic API key is already configured. Set a new value? (y/N): y
Enter your Anthropic API key (press Enter to skip): 
[INFO] Anthropic API key disabled
```

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -27,6 +27,10 @@ print_error() {
     echo -e "${RED}[ERROR]${NC} $1"
 }
 
+print_linebreak() {
+    echo "----------------------"
+}
+
 # Function to check if command exists
 command_exists() {
     command -v "$1" >/dev/null 2>&1
@@ -243,6 +247,7 @@ setup_config_file() {
 
 # Function to configure LangFuse
 configure_langfuse() {
+    print_linebreak
     print_status "Configuring LangFuse (optional monitoring)..."
     
     # Source the env file to check current values
@@ -476,22 +481,26 @@ configure_local_api_keys() {
     fi
     
     # OpenAI API Key (Optional)
+    print_linebreak
     print_status "OpenAI API Key (Optional): Powers AI-driven vulnerability analysis and patch generation."
     print_status "The patcher component performs best with OpenAI models (GPT-4o/GPT-4o-mini)."
     configure_service "OPENAI_API_KEY" "OpenAI API key" "$OPENAI_API_KEY" "<your-openai-api-key>" false
     
     # Anthropic API Key (Optional)
+    print_linebreak
     print_status "Anthropic API Key (Optional): Powers AI-driven fuzzing seed generation."
     print_status "The seed generation component performs best with Anthropic models (Claude 3.5/4 Sonnet)."
     configure_service "ANTHROPIC_API_KEY" "Anthropic API key" "$ANTHROPIC_API_KEY" "<your-anthropic-api-key>" false
     
     # GitHub Personal Access Token
-    print_status "GitHub Personal Access Token: Required to pull private competition containers."
+    print_linebreak
+    print_status "GitHub Personal Access Token: Required to access private resources."
     print_status "Local deployment required permissions: read challenge repos."
     print_status "Azure deployment required permissions: read GHCR packages and challenge repos."
     configure_service "GHCR_AUTH" "GitHub Container Registry authentication" "$GHCR_AUTH" "<your-ghcr-base64-auth>" true "configure_ghcr"
     
     # Docker Hub credentials (optional)
+    print_linebreak
     print_status "Docker Hub Credentials (Optional): Gives higher rate limits when pulling public base images."
     print_status "Recommended for reliable builds, but not strictly required for operation."
     configure_service "DOCKER_USERNAME" "Docker Hub credentials" "$DOCKER_USERNAME" "<your-docker-username>" false "configure_docker_hub_optional"
@@ -526,6 +535,7 @@ configure_local_api_keys() {
 
 # Function to configure OTEL telemetry
 configure_otel() {
+    print_linebreak
     print_status "Configuring OpenTelemetry telemetry (optional)..."
     
     # Source the env file to check current values


BradSwain:
I think we might need a better explanation for the GHCR stuff. Right now it says:

```
GitHub Container Registry: Required to pull private competition containers.
Your GitHub PAT needs 'package:read' permissions to access ghcr.io repositories.
Enter your GitHub username (press Enter to use 'USERNAME'): 
Enter your GitHub Personal Access Token (PAT): 
[SUCCESS] GitHub Container Registry authentication configured
```

But it may not be clear what private containers are or what specifically the PAT needs access to (or even how to create a ghcr PAT).

It looks like right now the PAT needs read access to `ghcr.io/aixcc-finals` but I'm not sure if this is something people outside of the competition will have access to.

Potentially conflicts with #164

reytchison:
> I think we might need a better explanation for the GHCR stuff. Right now it says:
> 
> ```
> GitHub Container Registry: Required to pull private competition containers.
> Your GitHub PAT needs 'package:read' permissions to access ghcr.io repositories.
> Enter your GitHub username (press Enter to use 'USERNAME'): 
> Enter your GitHub Personal Access Token (PAT): 
> [SUCCESS] GitHub Container Registry authentication configured
> ```
> 
> But it may not be clear what private containers are or what specifically the PAT needs access to (or even how to create a ghcr PAT).
> 
> It looks like right now the PAT needs read access to `ghcr.io/aixcc-finals` but I'm not sure if this is something people outside of the competition will have access to.
> 
> Potentially conflicts with #164

This is a good point, and even with my PR which is merged we could make it clearer. The PR explains that the read packages permission is only needed in an Azure deployment, but not what the packages are. The packages are the CRS components. In a local deployment, we don't pull the CRS images and we build them locally. Also, I think if we were using `ghcr.io/aixcc-finals` images then we might need the read packages permission for challenge containers? I tested a fine-grained PAT which only has repo permissions and it worked, presumably because we're using oss-fuzz containers now.

BradSwain:
> the read packages permission is only needed in an Azure deployment, but not what the packages are. The packages are the CRS components. In a local deployment, we don't pull the CRS images and we build them locally.

Nice! So if I understand correctly, in theory a local deployment should not need to provide GHCR auth. As long as all the images we build are based on public images (see below).

For an azure deployment are the components prebuilt and stored in GHCR? Would they be public after buttercup is released?

> Also, I think if we were using ghcr.io/aixcc-finals images then we might need the read packages permission for challenge containers?

I only see one image using aixcc-finals stuff (fuzzer uses  `ghcr.io/aixcc-finals/base-runner` as the base image).

https://github.com/trailofbits/buttercup/blob/913e51043c91be26f366e64ba01f1200dcdaca94/fuzzer/dockerfiles/runner_image.Dockerfile#L1

But there are references to `ghcr.io/aixcc-finals` in a few other places as well. e.g.

https://github.com/trailofbits/buttercup/blob/913e51043c91be26f366e64ba01f1200dcdaca94/deployment/k8s/charts/image-preloader/values.yaml#L16-L28

I was able to run `setup-local` though with a fine grained PAT only scoped to read the buttercup repo.

BradSwain:
The changes to allow the script to run multiple times are ready for review.

We may want to improve the descriptions and potentially make the PAT optional based on the above, but that can be done in another PR if needed.

reytchison:
> Nice! So if I understand correctly, in theory a local deployment should not need to provide GHCR auth. As long as all the images we build are based on public images (see below).

Yes, you made a good point that we still point to some aixcc base images. I was able to deploy the CRS with a fine-grained PAT that did not have GHCR permissions. I think this is because the AIxCC base images are now public: https://github.com/orgs/aixcc-finals/packages?visibility=public.

Also, for a minikube deployment we build images locally here: https://github.com/trailofbits/buttercup/blob/16405ef2d82f71090da1530bee3f0e63e6ffb531/deployment/crs-architecture.sh#L130-L136
And then we specify that we are using local images here:
https://github.com/trailofbits/buttercup/blob/16405ef2d82f71090da1530bee3f0e63e6ffb531/deployment/k8s/values-minikube.template#L1-L23
> For an azure deployment are the components prebuilt and stored in GHCR?

That's also correct. Also we're currently still pointing at the AIxCC GHCR:
https://github.com/trailofbits/buttercup/blob/16405ef2d82f71090da1530bee3f0e63e6ffb531/deployment/k8s/values.yaml#L7-L36
> Would they be public after buttercup is released?

I think so?

reytchison:
Update on the AIxCC base images for posterity. The base image for the fuzzer is overridden to be OSS-fuzz here: 
https://github.com/trailofbits/buttercup/blob/cb1dbd5f8af29b680618a39cb990dbe1b21813ce/deployment/env.template#L89
The image preloader values are overridden here: https://github.com/trailofbits/buttercup/blob/cb1dbd5f8af29b680618a39cb990dbe1b21813ce/deployment/k8s/values-upstream-minikube.template#L70-L81

Issue #175, opened 2025-07-25
Claude Code github action is broken


michaelbrownuc:
tried to use Claude to get #173 started and the workflow failed, needs someone to troubleshoot:

https://github.com/trailofbits/buttercup/actions/runs/16524969176

michaelbrownuc:
could be as simple as making `aixcc-cscope` public - job appears to fail when cloning a private submodule per Ronald

michaelbrownuc:
still broke

Smjert:
Reopening since it's still broken (although a different error): https://github.com/trailofbits/buttercup/actions/runs/16594129312/job/46936598554

`fatal: could not read Username for 'https://github.com/': No such device or address`

Smjert:
By the way I'm not sure this ever worked after adding the submodule and especially after setting `persist-credentials: false`. Basically, the couple of times it worked (looking at the workflow run history), it was using the credentials in the repository.

Smjert:
Closing again, since now this has been truly fixed by: https://github.com/trailofbits/buttercup/pull/198

PR #189, opened 2025-07-28
Fix claude code on CI

Smjert:
Restore clone of the source code since claude code depends on it.

Fixes https://github.com/trailofbits/buttercup/issues/175

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -24,6 +24,12 @@ jobs:
       issues: read
       id-token: write
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


Smjert:
@claude

Smjert:
Ah yeah, it won't use the workflow here. Will have to find another way to trigger hum.

Smjert:
I don't know that I can trigger Claude in doing more than this (https://github.com/trailofbits/buttercup/actions/runs/16572888302/job/46869683083), unfortunately only the workflow on the `main` branch is used when triggered by the phrase/word.

I'll remove the commit we don't want.

PR #198, opened 2025-07-29
Fix Claude Code action access to fetch the repo

Smjert:
Configure git executed by the Claude Code action to have access to the repo through env vars, to prevent persistence on disk of credentials.

The action implicitly assumes to have access to the repository when fetching but we are not persisting the credentials after cloning, so the action fails. While the action then retrieves a GITHUB_TOKEN on its own, this is not used by the initial fetch operation.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -31,6 +31,10 @@ jobs:
           persist-credentials: false
           submodules: true
       - name: Run Claude Code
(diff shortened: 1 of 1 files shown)


Smjert:
I was able to do a test using a separate branch/PR: https://github.com/trailofbits/buttercup/actions/runs/16599408766

Issue #169, opened 2025-07-24
Document what are the minimum and recommended hardware resources required to run


Smjert:
Especially for local development and or local evaluation, knowing how much disk space and ram are needed would be useful.
Even if it's not for installing everything on a laptop but sizing a VM correctly.

Smjert:
On this topic we might also want to preventively warn the user that, at least on macOS, they might have to increase the Docker Desktop allowed RAM usage (which I believe is around 8GB or so?).
This is somewhat minor because when doing `make deploy-local` you get a similar error anyway:
```
🧯  Requested memory allocation 32768MB is more than your system limit 32093MB.
```

But it could still be useful to get a smoother experience and to inform the user on what are the steps to change that.

disconnect3d:
We could also change the requested memory to be 8 or 16G or make it configurable/ask user, for better UX.

ret2libc:
We can do this, but things like memory and disk heavily depend on what projects you want to test the CRS on. If you want to do the integration-test challenge, that's one thing, if you want to build some big Java projects, you need other resources. Also, it depends on how many patches the system generate, how many vulns it finds, for how long you schedule each challenge, etc.

I agree we should have some sane defaults that would allow a random person to just test this on a smallish project, but if you want to run this for real you still need to thing about resources, similar to how we thought about them before deploying the CRS for the competition.

Also, the less resources, the less parallelization you can have. The system initially builds a few versions of a project with different sanitizers/coverage/etc. and if e.g. you have only one builder pod in the system (due to resource constraints, for example), it might take much longer for the system to even just start processing the project because the builds will happen sequentially.

Smjert:
I should've probably clarified that at least the initial intention is to document any example/default workload you have in the project, so that when a user initially tests the project they have something to quickly know what to prepare.

Simply when I cloned the project my only way to know that, as it was, it wanted 32GB of RAM, was to dig in the scripts or run it and have it fail, asking me for more RAM for Docker.

Then from there it would be ideal if we knew what are the fixed costs, and what depends on the project. My unfamiliarity with the project and architecture definitely played into it, but any guideline we can give without having the user dig through how everything is wired (and which software is used to do what etc) I think is important.

reytchison:
This PR documented the minimum specs and updated minikube to use them: https://github.com/trailofbits/buttercup/pull/245

michaelbrownuc:
completed

PR #201, opened 2025-07-29
scale minikube resources based on system

BradSwain:
Related #169

This changes the minikube deployment to use more resources, if available.

It also allows the deployment to run on systems with less than the previously hardcoded resources (8cpus, 32g mem).

It warns if resources are below some threshold, and errors if below a minimum threshold.

I took a best guess at what good values would be, but they can be changed/updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,9 @@ help:
 	@echo "  deploy-azure      - Deploy to production AKS environment"
 	@echo ""
 	@echo "Status:"
-	@echo "  status            - Check the status of the deployment"
+	@echo "  status              - Check the status of the deployment"
+	@echo "  crs-instance-id     - Get the CRS instance ID"
+	@echo "  download-artifacts  - Download submitted artifacts from the CRS"
 	@echo ""
 	@echo "Testing:"
 	@echo "  send-integration-task  - Run integration-test task"
@@ -116,6 +118,14 @@ status:
 	@echo "----------SERVICES--------"
 	@kubectl get services -n $${BUTTERCUP_NAMESPACE:-crs}
 	@make --no-print-directory check-crs
+
+download-artifacts:
+	@echo "Downloading artifacts from the CRS..."
+	@if ! kubectl get namespace $${BUTTERCUP_NAMESPACE:-crs} >/dev/null 2>&1; then \
+		echo "Error: CRS namespace not found. Deploy first with 'make deploy'."; \
+		exit 1; \
+	fi
+	./scripts/download_artifacts.sh
 
 # Testing targets
 send-integration-task:
--- a/deployment/k8s/charts/ui/templates/deployment.yaml
+++ b/deployment/k8s/charts/ui/templates/deployment.yaml
@@ -30,6 +30,8 @@ spec:
         volumeMounts:
           - name: data-volume
             mountPath: /data
+          - name: run-data-volume
+            mountPath: /tmp/buttercup-run-data
         command: ["buttercup-ui"]
         resources:
           {{- toYaml .Values.resources | nindent 12 }}
@@ -48,4 +50,13 @@ spec:
       volumes:
         - name: data-volume
           emptyDir: {}
+        {{- if eq .Values.global.environment "minikube" }}
+        - name: run-data-volume
+          hostPath:
+            path: /data/{{ .Release.Name }}-run-data
+            type: DirectoryOrCreate
+        {{- else }}
+        - name: run-data-volume
+          emptyDir: {}
+        {{- end }}
 {{- end }} 
--- a/deployment/k8s/templates/common-env.yaml
+++ b/deployment/k8s/templates/common-env.yaml
@@ -364,6 +364,8 @@ Service-specific environment variables that utilize the standardized variables
   value: {{ .Values.global.crs.api_key_token | quote }}
 - name: BUTTERCUP_UI_STORAGE_DIR
   value: "/data/challenges"
+- name: BUTTERCUP_UI_RUN_DATA_DIR
+  value: "/tmp/buttercup-run-data"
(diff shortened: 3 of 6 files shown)


deployment/crs-architecture.sh
```diff
@@ -94,7 +94,46 @@ up() {
...
 				echo -e "${BLU}Deploying minikube cluster${NC}"
-				minikube status | grep -q "Running" || minikube start --force --extra-config=kubeadm.skip-phases=preflight --cpus=8 --memory=32g --disk-size=80g --driver=docker --kubernetes-version=stable
+				
+				# Detect system resources and adjust minikube parameters
+				if [[ "$OSTYPE" == "darwin"* ]]; then
+					# macOS
+					AVAILABLE_CPUS=$(sysctl -n hw.ncpu)
+					AVAILABLE_MEMORY_BYTES=$(sysctl -n hw.memsize)
+					AVAILABLE_MEMORY_MB=$((AVAILABLE_MEMORY_BYTES / 1024 / 1024))
+				else
```
ret2libc: at least in the case of darwin i think this depends more on what you assigned to colima/docker desktop rather than the system itself.

ret2libc: maybe we can just use `docker system info` data and get the cpus/memory from there?

BradSwain:
@claude  use docker system info data and get the cpus/memory from there

deployment/crs-architecture.sh
```diff
@@ -94,7 +94,46 @@ up() {
...
+				AVAILABLE_CPUS=$(echo "$DOCKER_INFO" | cut -d',' -f1)
+				AVAILABLE_MEMORY_BYTES=$(echo "$DOCKER_INFO" | cut -d',' -f2)
+				AVAILABLE_MEMORY_MB=$((AVAILABLE_MEMORY_BYTES / 1024 / 1024))
+				
+				# Calculate safe limits (leave 25% for system)
+				MINIKUBE_CPUS=$((AVAILABLE_CPUS > 7 ? 8 : AVAILABLE_CPUS > 4 ? AVAILABLE_CPUS - 1 : AVAILABLE_CPUS))
+				MINIKUBE_MEMORY_MB=$((AVAILABLE_MEMORY_MB * 75 / 100))
+				
+				# Check if system meets minimum requirements
+				if [ $AVAILABLE_CPUS -lt 2 ]; then
```
ret2libc: i think while looking at nodes usage, i saw that the system with just one build-bot was using a bit more than 4 cpus, so i guess the minimum should be 5

deployment/crs-architecture.sh
```diff
@@ -94,7 +94,46 @@ up() {
...
+					echo -e "${RED}Error: System has only ${AVAILABLE_CPUS} CPU(s), but Buttercup requires at least 2 CPUs${NC}"
+					echo -e "${RED}Please use a machine with more CPU cores${NC}"
+					exit 1
+				fi
+				
+				AVAILABLE_MEMORY_GB=$((AVAILABLE_MEMORY_MB / 1024))
+				if [ $AVAILABLE_MEMORY_MB -lt 16384 ]; then  # Need 16GB total for Buttercup components
```
ret2libc: i think the minimum for buttercup to work is around 9/10 gb of ram dedicate to the k8s cluster.

ret2libc:
@claude address the review

deployment/crs-architecture.sh
```diff
@@ -94,7 +94,46 @@ up() {
...
+					echo -e "${RED}Please use a machine with more memory${NC}"
+					exit 1
+				fi
+				
+				
+				MINIKUBE_MEMORY_GB=$((MINIKUBE_MEMORY_MB / 1024))
+				
+				echo -e "${GRN}Docker-available resources: ${AVAILABLE_CPUS} CPUs, ${AVAILABLE_MEMORY_GB}GB RAM${NC}"
+				echo -e "${GRN}Minikube will use: ${MINIKUBE_CPUS} CPUs, ${MINIKUBE_MEMORY_GB}GB RAM${NC}"
+				echo -e "${BLU}Note: Recommended minimum is at least 8 CPUs and 64GB RAM${NC}"
```
ret2libc:
```suggestion
				echo -e "${BLU}Note: Recommended minimum is at least 5 CPUs and 10GB RAM${NC}"
```

deployment/crs-architecture.sh
```diff
@@ -87,7 +87,46 @@ up() {
...
+				DOCKER_INFO=$(docker system info --format "{{.NCPU}},{{.MemTotal}}")
+				AVAILABLE_CPUS=$(echo "$DOCKER_INFO" | cut -d',' -f1)
+				AVAILABLE_MEMORY_BYTES=$(echo "$DOCKER_INFO" | cut -d',' -f2)
+				AVAILABLE_MEMORY_MB=$((AVAILABLE_MEMORY_BYTES / 1024 / 1024))
+				
+				# Calculate safe limits (leave 25% for system)
+				MINIKUBE_CPUS=$((AVAILABLE_CPUS > 7 ? 8 : AVAILABLE_CPUS > 4 ? AVAILABLE_CPUS - 1 : AVAILABLE_CPUS))
+				MINIKUBE_MEMORY_MB=$((AVAILABLE_MEMORY_MB * 75 / 100))
+				
+				# Check if system meets minimum requirements
```
reytchison, line 108: The README should document the minimum requirements

reytchison: This could also go in the #214 PR

BradSwain:
Got pulled to some other work, but I plan to make the resources configurable during setup instead of during deployment.

I made env vars to configure minikube resources, pod count, and individual pod resource, and moved the "detect resources" logic to a script that runs during local setup

BradSwain:
Closed in favor of listing the minimum required resources in #214

PR #202, opened 2025-07-30
Make CRS more flexible

ret2libc:
I tried to make the system work on the integration-test challenge, but unfortunately it still doesn't because that challenge has 2 vulnerabilities in the same function (essentially, they are considered the same vuln), so a proper patch would need to fix both of them. However, to fix one of those you also have to patch the fuzzing harness, which is not possible because of AIxCC constraints and how the system was designed, so the patcher can't create a patch.

I'll probably switch to just run the aixcc delta libpng version for quick testing and showcasing the CRS, so it's still small but it should work.


reytchison:
Noting that the base image must still be configured for the Fuzzer base image:
https://github.com/trailofbits/buttercup/blob/4e4c173a24944fdc3f90bd643c58b03818b64ade/deployment/env.template#L84

Issue #149, opened 2025-07-09
Adapt Buttercup to ARM


michaelbrownuc:
Buttercup currently has only been tested on x86 targets and only runs on x86 coder instances. To run on Mac hardware natively we need to do:

- [ ] create exemplar C/C++/Java programs that can build / run on ARM
- [ ] test Buttercup components with these exemplars
- [ ] fix issues resulting from testing

ret2libc:
I think x86_64 is hardcoded throughout the codebase right now.

ret2libc:
https://github.com/trailofbits/buttercup/blob/cdb67b0b60b984dd25c12b0b7b116971e7df2c62/orchestrator/src/buttercup/orchestrator/scheduler/submissions.py#L272

ret2libc:
```diff
diff --git a/fuzzer/dockerfiles/runner_image.Dockerfile b/fuzzer/dockerfiles/runner_image.Dockerfile
index 488fcb9..340125e 100644
--- a/fuzzer/dockerfiles/runner_image.Dockerfile
+++ b/fuzzer/dockerfiles/runner_image.Dockerfile
@@ -11,7 +11,7 @@ ENV UV_PYTHON_DOWNLOADS=manual
 RUN uv python install python3.10

 FROM base-image AS runner-base
-RUN apt-get update
+RUN apt-get update && apt-get install -y curl
 # TODO(Ian): maybe we should have a different base image for the builder
 RUN curl -fsSL https://get.docker.com | sh
```

This is apparently necessary on the ARM image

PR #207, opened 2025-07-30
Add support for ARM

ret2libc:
based on #202 , need to first merge that and then update the base branch of this PR.

----

* detect architecture from running machine and use that instead of hardcoding x86_64.
* pass the architecture to all oss-fuzz related commands, because there’s an --architecture flag in oss-fuzz to build for arm
* fix fuzzer docker and common script (stolen from Evan’s branch)
* modified our own fork of example-crs-architecture to extend the API so that it accepts also aarch64 as architecture. Without this change, the APIs would only allow `x86_64`.
* used libucl as the example challenge, because it seems to find issues very very quickly, it’s public and upstream, it’s small enough.
* fixed program-model docker so that it doesn’t hardcode the SHA of the x86_64 container base image

Closes https://github.com/trailofbits/buttercup/issues/149

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ lint-component:
 	fi
 	@echo "Linting $(COMPONENT)..."
 	@cd $(COMPONENT) && uv sync -q --all-extras && uv run ruff format --check && uv run ruff check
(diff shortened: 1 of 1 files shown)


ret2libc:
- [ ] Add `Co-authored-by: Evan Downing <[2077950+evandowning@users.noreply.github.com](mailto:2077950+evandowning@users.noreply.github.com)>`

PR #210, opened 2025-07-31
Remove just tool completely and migrate all functionality to Makefiles

### LINKED ISSUE #173: Remove `just` tool
while I typically prefer to use `just`, recent PRs have consolidated much of the system automation into the `Makefile`.  Right now the `justfile` has virtually no functions and can be moved to the `Makefile`. We can also skip installing it on the user's system.

ret2libc:
Fixes #173

This PR completely removes the `just` tool from the repository and migrates all its functionality to standard Makefiles.

## Changes
- Added missing targets to main Makefile: `install-cscope`, `lint-python`, `lint-python-all`
- Created program-model/Makefile with all justfile functionality (all, sync, reformat, lint, test)
- Removed justfiles from both main and program-model directories
- Updated all documentation to use make commands instead of just
- Removed just installation and checking functions from setup scripts

## Benefits
- No longer requires users to install the `just` tool
- Uses standard `make` which is available on most systems by default
- All existing functionality preserved with equivalent make targets

### DIFF
--- a/.github/ci-env.template
+++ b/.github/ci-env.template
@@ -26,4 +26,3 @@ export LANGFUSE_PUBLIC_KEY=replace-me
 export LANGFUSE_SECRET_KEY=replace-me
 export DOCKER_USERNAME=replace-me
 export DOCKER_PAT=replace-me
-export FUZZ_TOOLING_CONTAINER_ORG=replace-me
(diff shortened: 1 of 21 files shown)


ret2libc:
@claude remove lint-python and lint-python-component targets from Makefile as they are already implemented as `lint` and `lint-component`.

ret2libc:
@claude fix the CI

ret2libc:
@claude fix the CI. probably there are still some references to `just` in the `.github` workflows. Fix them.

Issue #211, opened 2025-07-31
buttercup-ui: implement a GUI to show the current status of the system


ret2libc:
Buttercup-ui should be extended to provide a service (also exposed in k8s) with a GUI that can be used to submit new tasks to the CRS and to view a dashboard of what the CRS has done so far and what it is working on.

- have a way to submit a new task given some info. You can check the info required to submit a new task by looking at ./orchestrator/scripts/task_crs.sh. Those fields should be requested to the user.
- have a dashboard that shows how many tasks the CRS is currently working on. Tasks have a deadline and are valid only until the deadline. Buttercup-ui should know when a task is submitted and its deadline, so it should be easy for it to know when a task is active and when not
- in the dashboard we want to see how many PoVs have been submitted by the CRS in total, how many patches, how many bundles.
- there should be a way to see task details: in that case, you want to see povs/bundles/patches only for that specific task
- there should be a way to see a PoV detail, a bundle detail, and a patch detail. PoVs are binary, so be careful with what we show. Maybe hexdump? or just base64 encoded or show only if small and not binary data. Find how the best way to do this. For patches, they can be long sometimes, but they are only text so it should be good to show them.
- a task has a name (the oss-fuzz project name), a task id, and a few other info. show those somehow so that users can easily determine what's the task about.

The dashboard should be auto-updated in real time, so users can see what's going on.

ret2libc:
@claude work on this. Base your branch on top of #206

PR #212, opened 2025-07-31
feat: implement GUI dashboard for buttercup-ui

ret2libc:
Implements issue #211 - GUI dashboard for Buttercup CRS system

Adds a comprehensive web dashboard with:
- Task submission form with all required fields
- Real-time system status and statistics
- Task listing with filtering capabilities
- Detailed task views showing PoVs, patches, bundles
- Safe binary data display for PoVs
- Responsive design with auto-refresh

image
image
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,10 @@ validate:
 # Deployment targets
 deploy:
 	@echo "Deploying to current environment..."
+	@if [ ! -f external/aixcc-cscope/configure.ac ]; then \
+		echo "Error: The git submodules have not been initialized. Run 'git submodule update --init --recursive' first."; \
+		exit 1; \
+	fi
 	cd deployment && make up
 	make wait-crs
 
@@ -97,6 +101,10 @@ deploy-local:
 		echo "Error: Configuration file not found. Run 'make setup-local' first."; \
 		exit 1; \
 	fi
+	@if [ ! -f external/aixcc-cscope/configure.ac ]; then \
+		echo "Error: The git submodules have not been initialized. Run 'git submodule update --init --recursive' first."; \
+		exit 1; \
+	fi
 	cd deployment && make up
 	make crs-instance-id
 	make wait-crs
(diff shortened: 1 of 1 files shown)


ret2libc:
@claude 
- default task duration should be 30mins
- fuzz tooling url should be prefilled/default with https://github.com/google/oss-fuzz and `master` ref
- "harnesses included" should default to true (actually, better to disable it completely, since we do not support unharnessed challenges for now)
- the placeholders for the challenge and fuzz-tooling URLs should be https://.. and not git@....
- remove the "completed" status. there's never a "completed" status. it's either active, or expired.

ret2libc:
@claude 
- add a progress-bar or spinning or something after pressing the "Submit task" button because it can take a bit before the task is submitted to the CRS.
- when pressing on a pov/bundle/patch in the task-detail, it should move to a new page where it shows the detail of the pov/bundle/patch. In this page there should also be a button to download the pov/patch/bundle (binary data for pov/patch, just json for the bundle)
- there should be a few other tabs "PoVs" and "Patches" where you can see the list of all povs/patches across all tasks. Again, if you press on any of these you should see the details.

ret2libc:
@claude
- patch download fails, fix it
- both the patch and the pov are showed in base64, but that's not very useful... it would be better to see the first few lines of an hexdump, for example. or something like that. also, the patch shhould be shown decoded, not as base64.
- when clicking on the dashboard buttons "Total POVs" and "Total patches", it should go to the right tab for that.

ret2libc:
@claude

- pov hexdump shown in the "POVs" tab are not displayed as well as in the task detail page. The image show what i'd like to see also in the POV detail page.

image

- when clicling on "Active tasks" it should select the "Tasks" tab.

ret2libc:
@claude for tasks, also show when it was created. There's "deadline" right now, but it would be nice to also see the "created_at" field.

ret2libc:
@claude 
- i'd like the detail panes to be a bit larger, so that more info can fit in.
- the "Download PATCH" button is still failing. It says "download failed"

ret2libc:
@claude the patch and bundle downloads are still failing!

ret2libc:
@claude merge the main branch into this one and handle the conflicts. the main difference is that povs/patches are also saved on disk and that functionality should be preserved.

ret2libc:
@claude add another button for "Submit example-libpng task" near the "Submit task" button. This new button should submit a default task to the CRS.

```
#!/bin/bash
    "fuzz_tooling_ref": "master",
    "duration": 1800
}'
```
see this to understand the values.

PR #216, opened 2025-07-31
scripts: Convert the challenge.sh script to Python

alessandrogario:
Latest macOS still relies on bash 3.2.57, which is not compatible with the constructs used by the challenge.sh script.

### DIFF
--- a/orchestrator/scripts/challenge.py
+++ b/orchestrator/scripts/challenge.py
@@ -594,7 +594,7 @@ name = The name of the challenge to run:
     testing_parser.set_defaults(func=lambda args: testing())
 
     single_parser = subparsers.add_parser("single", help="Run one challenge")
-    single_parser.add_argument("name", type=str, help="Challenge name")
+    single_parser.add_argument("name", type=str, choices=list(CHALLENGE_MAP.keys()), help="Challenge name")
     single_parser.add_argument("duration", type=int, help="Duration in seconds")
     single_parser.set_defaults(func=lambda args: single(args.name, args.duration))
 


orchestrator/scripts/challenge.py
ret2libc, line 1: is this executable?

orchestrator/scripts/challenge.py
```diff
@@ -0,0 +1,615 @@
...
+import urllib.request
+import sys
+from typing import Any
+
+SECONDS: int = 1
+MINUTES: int = 60 * SECONDS
+HOURS: int = 60 * MINUTES
+
+FULL_SET_DURATION: int = 24 * HOURS
+DELTA_SET_DURATION: int = 8 * HOURS
```
ret2libc: again, these types are really not necessary and usually make the code more verbose. They don't hurt, but they are not that necessary in python.

orchestrator/scripts/challenge.py
```diff
@@ -0,0 +1,615 @@
...
+
+            except (json.JSONDecodeError, UnicodeDecodeError):
+                print(f"{req.data}", file=file)
+        else:
+            print(f"{req.data}", file=file)
+
+    print("\n")
+
+
+def submit_task(task_name: str, overrides: list[tuple[str, Any]] = []) -> None:
```
ret2libc:
```suggestion
def submit_task(task_name: str, **kwargs: Any) -> None:
```

i think it's just better if you allow for kwargs are use that as overrides. Then, when you call this method you do:
```python
submit_task("cc_full_01", duration=full_set_duration)
```
instead of:
```python
submit_task("cc_full_01", [("duration", full_set_duration)])
```

orchestrator/scripts/challenge.py
alessandrogario: Yes, and has the `#!/usr/bin/env python3` hashbang

orchestrator/scripts/challenge.py
alessandrogario: Oops, habit from Rust global consts!

orchestrator/scripts/challenge.py
```diff
@@ -0,0 +1,615 @@
...
+                data_str = req.data.decode("utf-8")
+                data_json = json.loads(data_str)
+                print(json.dumps(data_json, indent=2), file=file)
+
+            except (json.JSONDecodeError, UnicodeDecodeError):
+                print(f"{req.data}", file=file)
+        else:
+            print(f"{req.data}", file=file)
+
+    print("\n")
```
ret2libc:
```suggestion
    print("\n", file=file)
```

orchestrator/scripts/challenge.py
ret2libc, line 1: use `argparse`

PR #218, opened 2025-08-01
Auto set resources on `make setup-local`

BradSwain:
Potentially replaces #201

Makes the resources configurable values set in the `env` file. The values are automatically set during `make setup-local` based on available system resources.

TODOS:
 - [ ] set minimums that make sense
 - [ ] allocate resources based experience with sample challenges
 - [ ] (optional) Do not overwrite manually configured values when running `make setup-local`

@ret2libc What do you think of this approach in general compared to #201 ?

### DIFF
--- a/deployment/k8s/values-minikube.template
+++ b/deployment/k8s/values-minikube.template
@@ -46,54 +46,6 @@ redis:
   master:
     persistence:
       enabled: false
-    resources:
-      limits:
-        cpu: ${REDIS_CPU_LIMIT}
-        memory: ${REDIS_MEMORY_LIMIT}
-      requests:
-        cpu: ${REDIS_CPU_REQUEST}
-        memory: ${REDIS_MEMORY_REQUEST}
-
-# Service-specific configurations with dynamic resource allocation
-build-bot:
-  replicaCount: ${BUILD_BOT_REPLICAS}
-  resources:
-    limits:
-      cpu: ${BUILD_BOT_CPU_LIMIT}
-      memory: ${BUILD_BOT_MEMORY_LIMIT}
-    requests:
-      cpu: ${BUILD_BOT_CPU_REQUEST}
-      memory: ${BUILD_BOT_MEMORY_REQUEST}
-
-scheduler:
-  resources:
-    limits:
-      cpu: ${SCHEDULER_CPU_LIMIT}
-      memory: ${SCHEDULER_MEMORY_LIMIT}
-    requests:
-      cpu: ${SCHEDULER_CPU_REQUEST}
-      memory: ${SCHEDULER_MEMORY_REQUEST}
-
-dind-daemon:
-  resources:
-    limits:
-      cpu: ${DIND_CPU_LIMIT}
-      memory: ${DIND_MEMORY_LIMIT}
-    requests:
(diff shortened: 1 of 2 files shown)


deployment/crs-architecture.sh
ret2libc, line 1: set defaults to MINIKUBE_* vars if not present. On a production deployment you probably don't evan care about setting those vars in the env file

deployment/env.template
```diff
@@ -82,3 +82,39 @@ export ANTHROPIC_API_KEY="<your-anthropic-api-key>"
 
 # Docker build arguments, useful for local deployment
 export FUZZER_BASE_IMAGE="gcr.io/oss-fuzz-base/base-runner"
+
+# Minikube cluster resource allocation
```
ret2libc, line 86: Explain a bit more how/when these are used, saying they are for the minikube environment only. Also, maybe make these commented out by default and enable them in the detect-resources script?! WDYT?

scripts/detect-resources.sh
```diff
@@ -0,0 +1,384 @@
...
+            if [ "$total_memory_bytes" != "0" ] && [[ "$total_memory_bytes" =~ ^[0-9.]+$ ]]; then
+                total_memory_bytes=$(echo "$total_memory_bytes * 1073741824" | bc 2>/dev/null || echo "0")
+            else
+                total_memory_bytes=0
+            fi
+        fi
+    fi
+    
+    # Convert memory from bytes to MiB
+    local total_memory_mib=$((total_memory_bytes / 1048576))
```
ret2libc, line 85: shouldn't this be done somehow by the `convert_to_mib` function?

scripts/detect-resources.sh
```diff
@@ -0,0 +1,384 @@
...
+                ;;
+            *)
+                print_warning "Unsupported platform: $platform, using default values"
+                ;;
+        esac
+    fi
+    
+    # Set minimum values if detection failed
+    [ "$total_cpus" = "0" ] && total_cpus=4
+    [ "$total_memory_mib" = "0" ] && total_memory_mib=4096
```
ret2libc, line 126:
```suggestion
    [ "$total_cpus" = "0" ] && total_cpus=5
    [ "$total_memory_mib" = "0" ] && total_memory_mib=9216
```

scripts/detect-resources.sh
ret2libc: and actually that fnction is never used, I think

michaelbrownuc:
> overall I think it's a nice idea, however I'm not convinced all this logic is really needed. We are talking anyway about a very test-based deployment, just for showing the program. For _any real usage_, people will have to play a bit with the pods allocation themselves and give the right resources. For example, just having one seed-gen and one fuzzer-bot is not going to work very well. Mid/long-term i think we are better of implementing auto-scaling within k8s so that we can start everything with just 1 pod per component and scale up/down automatically as needed.
> 
> The reason why I think this might be a bit over-engineered is that given it's just for testing/showing the tool, ensuring the system has a minimum of X cpu/Y mem is enough to deploy a basic system that works. Any other use case above that will require some thinking about resources allocated to each component.
> 
> cc @michaelbrownuc

I agree here, this is nice to have but it's probably not the most pressing priority.

PR #221, opened 2025-08-01
Add submit-project script for custom challenge submission

### LINKED ISSUE #220: Create a `submit-project` script/make-target to guide the user through submitting a custom challenge
This should be a python script, most likely, that ask the user a bunch of questions to determine how to submit the task to the buttercup-ui.

IOW, the script will end up doing a request like this:
```
#!/bin/bash
    "fuzz_tooling_ref": "master",
    "harnesses_included": true,
    "duration": 1800
}'
```

and it needs to determine the various values for the fields.
Some questions that might be asked to the user:
- Do you want to use upstream google/oss-fuzz? Y/n
  - if not, the custom oss-fuzz url/ref should be asked
- what's the oss-fuzz project name?
- do you want to analyze a specific branch/commit? y/N
  - if yes, we should ask for the branch/commit. if there's `..` in the user's input, it should be treated as a diff mode, otherwise a full mode challenge
  - if no (the default), the script should determine the default branch name and use that in the head_ref field.
- how long should the analysis last? (in minutes)

ret2libc:
Add interactive Python script for custom challenge submission

- Created interactive Python script in scripts/submit-project.py
- Guides users through submitting challenges via prompts
- Auto-detects default branch from git repositories
- Supports diff mode (branch1..branch2) and full mode analysis
- Validates repository URLs before submission
- Includes dry-run mode for testing
- Added make target 'submit-project' with port forwarding
- Handles upstream/custom oss-fuzz configurations
- Proper error handling and user-friendly messages

Resolves #220

### DIFF
--- a/scripts/submit-project.py
+++ b/scripts/submit-project.py
@@ -15,6 +15,12 @@ import urllib.parse
 from pathlib import Path
 from typing import Optional
 
+try:
+    import yaml
+    HAS_YAML = True
+except ImportError:
+    HAS_YAML = False
+
 
 def run_git_command(cmd: list[str], cwd: Optional[Path] = None) -> str:
     """Run a git command and return its output."""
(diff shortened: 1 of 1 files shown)


scripts/submit-project.py
```diff
@@ -0,0 +1,282 @@
...
+        fuzz_tooling_ref = "master"
+    else:
+        fuzz_tooling_url = ask_string("Custom oss-fuzz repository URL")
+        fuzz_tooling_ref = ask_string("OSS-Fuzz repository reference", "master")
+    
+    fuzz_tooling_project_name = ask_string("OSS-Fuzz project name")
+    
+    # Harnesses
+    harnesses_included = ask_yes_no("Are harnesses included in the project?", True)
+
```
ret2libc: this should be always True and non-configurable

ret2libc:
@claude

ret2libc:
@claude address the review

ret2libc:
Maybe better to extend #216 later on

PR #229, opened 2025-08-01
Enforce mypy on fuzzer component

### LINKED ISSUE #224: Enforce mypy on fuzzer component

BradSwain:
Fixes #224

This PR enforces mypy type checking on the fuzzer component by:

- Adding fuzzer to the mypy enforcement list in Makefile
- Fixing 39 type errors across 9 files in the fuzzer component
- Adding missing return type annotations
- Fixing Beautiful Soup union type issues
- Handling TemporaryDirectory path usage correctly
- Adding proper None checks and type casting

All fuzzer component files now pass strict mypy type checking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Choose your deployment method:
 - **[Local Development](#local-development)** - Quick setup for development and testing
 - **[Production AKS Deployment](#production-aks-deployment)** - Full production deployment on Azure Kubernetes Service
 
+Note: Buttercup uses hosted LLMs, which cost money. Limit your per-deployment spend with the built-in LLM budget.
+
 ## Local Development
 
 The fastest way to get started with the **Buttercup CRS** system for development and testing.
--- a/deployment/env.template
+++ b/deployment/env.template
@@ -59,6 +59,7 @@ export GHCR_AUTH="<your-ghcr-base64-auth>"
 
 # LiteLLM/LLMs settings
 export LITELLM_MASTER_KEY="d5179c62ae1c7366e3ee09775d0993d5" # Generate with: openssl rand -hex 16
+export LITELLM_MAX_BUDGET="100"
 export AZURE_API_BASE="<disabled>"
 export AZURE_API_KEY="<disabled>"
 export OPENAI_API_KEY="<your-openai-api-key>"
--- a/deployment/k8s/charts/patcher/templates/deployment.yaml
+++ b/deployment/k8s/charts/patcher/templates/deployment.yaml
@@ -38,6 +38,7 @@ spec:
         {{- include "buttercup.patcherEnv" . | nindent 8 }}
         {{- include "buttercup.env.dockerSocket" . | nindent 8 }}
         {{- include "buttercup.env.langfuse" . | nindent 8 }}
+        {{- include "buttercup.env.llm" (merge (dict "secretName" "litellm-api-user" "secretKey" "API_KEY") .) | nindent 8 }}
         - name: LOG_LEVEL
           value: "{{ .Values.logLevel }}"
       volumes:
@@ -45,4 +46,5 @@ spec:
       {{- include "buttercup.volumes.tasks" . | nindent 6 }}
       {{- include "buttercup.dockerSocketVolume" . | nindent 6 }}
       {{- include "buttercup.nodeLocalVolume" . | nindent 6 }}
+      {{- include "buttercup.apiKeySecretVolume" (dict "secretName" "litellm-api-user") | nindent 6 }}
 {{- end }}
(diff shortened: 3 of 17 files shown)


fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
```diff
@@ -35,7 +35,7 @@ class BuilderBot:
     _build_outputs_queue: ReliableQueue[BuildOutput] = field(init=False)
     _registry: TaskRegistry | None = field(init=False, default=None)
```
ret2libc:
```suggestion
    _registry: TaskRegistry = field(init=False)
```

fuzzer/src/buttercup/fuzzing_infra/builder_bot.py
```diff
@@ -107,10 +107,10 @@ def serve_item(self) -> bool:
         )
 
         # Check if task should not be processed (expired or cancelled)
-        if self._registry.should_stop_processing(msg.task_id):
+        if self._registry is not None and self._registry.should_stop_processing(msg.task_id):
```
ret2libc:
```suggestion
        if self._registry.should_stop_processing(msg.task_id):
```

fuzzer/src/buttercup/fuzzing_infra/corpus_merger.py
```diff
@@ -97,7 +97,7 @@ def __post_init__(self):
 
         for file in local_files_list:
             try:
-                shutil.copy(os.path.join(self.corpus.path, file), os.path.join(self.local_dir, file))
+                shutil.copy(os.path.join(self.corpus.path, file), os.path.join(self.local_dir.name, file))
```
ret2libc: the node_local types are a bit messed up tbh, but i think this is the wrong adjustments... probably the type of `local_dir` and `remote_dir` should be just `str` or `Path`... needs to be investigated manually.

fuzzer/src/buttercup/fuzzing_infra/corpus_merger.py
```diff
@@ -111,10 +111,10 @@ def __post_init__(self):
 
         for file in self.remote_files:
             try:
-                shutil.copy(os.path.join(self.corpus.path, file), os.path.join(self.remote_dir, file))
+                shutil.copy(os.path.join(self.corpus.path, file), os.path.join(self.remote_dir.name, file))
```
ret2libc: same for this and the others.

fuzzer/src/buttercup/fuzzing_infra/corpus_merger.py
```diff
@@ -205,7 +205,16 @@ def __init__(
     def required_builds(self) -> List[BuildTypeHint]:
         return [BuildType.FUZZER]
 
-    def _run_merge_operation(self, task, build, remote_dir, local_dir, local_only_files, remote_files, corp: Corpus):
+    def _run_merge_operation(
+        self,
+        task: WeightedHarness,
+        build: BuildOutput,
+        remote_dir: Any,
+        local_dir: Any,
```
ret2libc: also these are probably just str or Path

fuzzer/tests/test_merger_bot.py
ret2libc, line 1: review these tests after changing the types of local_dir/remote_dir

fuzzer/src/buttercup/fuzzing_infra/corpus_merger.py
BradSwain: If I understand correctly, they are `TmpDir`s

https://github.com/trailofbits/buttercup/blob/7eb017a6ad8b139399e06816a57eaf7e00c3807f/common/src/buttercup/common/node_local.py#L26-L32

Which looks to be implementing [`os.PathLike`][1]:


> class os.PathLike[¶](https://docs.python.org/3/library/os.html#os.PathLike)
An [abstract base class](https://docs.python.org/3/glossary.html#term-abstract-base-class) for objects representing a file system path, e.g. [pathlib.PurePath](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath).
>
> abstractmethod __fspath__()
>Return the file system path representation of the object.
>
> The method should only return a [str](https://docs.python.org/3/library/stdtypes.html#str) or [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) object, with the preference being for [str](https://docs.python.org/3/library/stdtypes.html#str).


So I made them `Pathlike[str]`

[1]: https://docs.python.org/3/library/os.html#os.PathLike

michaelbrownuc:
merging, CI is back to green

PR #232, opened 2025-08-04
feat: Add local SigNoz deployment support

### LINKED ISSUE #226: Deploy a local signoz and log there by default
In `deployment/` we should add a SigNoz service so that when deploying the system locally SigNoz is also installed and users can monitor the system run without having access to an external SigNoz service.

The `values-upstream-minikube` template file should be changed to have something like `signoz: enabled: true`, while it should be disabled by default.

In `common-env` file the OTEL endpoint should be set to that internal SigNoz service by default if not specified otherwise.

ret2libc:
This PR implements a minimal SigNoz integration using the official Helm chart, as requested in issue #226.

## Changes
- Add SigNoz as a Helm dependency with conditional deployment
- Enable SigNoz by default for minikube environments
- Auto-configure OTEL to use internal SigNoz when enabled
- Update Docker Compose to include existing SigNoz stack
- Minimal configuration following the official chart approach

Closes #226

### DIFF
--- a/CUSTOM_CHALLENGES.md
+++ b/CUSTOM_CHALLENGES.md
@@ -1,0 +1,376 @@
+# Writing Custom Challenges
+
+Buttercup is an automated cyber reasoning system originally developed for DARPA's [AI Cyber Challenge][1]. While it was designed to work seamlessly with existing oss-fuzz projects, you can also configure it to analyze custom projects with some additional setup.
+
+This guide walks you through creating custom challenges for Buttercup, covering everything from the basic input format to debugging common issues.
+
+## Prerequisites
+
+To analyze a custom project, you'll need to provide:
+
+- **Fuzzing harnesses** that serve as entry points for analysis
+- **Build configuration** in a forked oss-fuzz repository
+- **Project metadata** defining how to compile and run your fuzzers
+
+## Challenge Input Format
+
+Buttercup accepts challenges through a JSON configuration that specifies both the target project and the fuzzing infrastructure. Here's a complete example:
+
+```json
+{
+    "challenge_repo_url": "https://github.com/tob-challenges/integration-test.git",
+    "challenge_repo_head_ref": "challenges/integration-test-delta-01",
+    "fuzz_tooling_url": "https://github.com/tob-challenges/oss-fuzz-aixcc",
+    "fuzz_tooling_ref": "challenge-state/integration-test-delta-01",
+    "fuzz_tooling_project_name": "integration-test",
+    "duration": 1800,
+}
+```
+
+### Field Descriptions
+
+**Project Repository Fields:**
+- `challenge_repo_url`: URL to the repository containing your source code
+- `challenge_repo_head_ref`: Git branch, tag, or commit to analyze
+- `challenge_repo_base_ref`: *(Delta mode only)* The "clean" commit before vulnerabilities were introduced
+
+**Fuzzing Infrastructure Fields:**
+- `fuzz_tooling_url`: URL to your forked oss-fuzz repository with build instructions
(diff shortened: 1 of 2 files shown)


deployment/k8s/templates/common-env.yaml
```diff
@@ -85,12 +85,21 @@ Define Docker Host environment variable for Unix socket
 {{- end }}
 
 {{- define "buttercup.env.telemetry" }}
+{{- if .Values.global.signoz.deployed }}
+- name: OTEL_EXPORTER_OTLP_ENDPOINT
+  value: "http://{{ .Release.Name }}-signoz-otel-collector:4317"
+- name: OTEL_EXPORTER_OTLP_HEADERS
+  value: "Authorization=Basic dXNlcm5hbWU6cGFzc3dvcmQ="
```
BradSwain, line 92: Is the authorization intended to be hardcoded?

Makefile
```diff
@@ -213,6 +214,22 @@ clean-local:
 install-cscope:
 	cd external/aixcc-cscope/ && autoreconf -i -s && ./configure && make && sudo make install
 
+signoz-ui:
+	@echo "Opening SigNoz UI..."
+	@if ! kubectl get namespace $${BUTTERCUP_NAMESPACE:-crs} >/dev/null 2>&1; then \
+		echo "Error: CRS namespace not found. Deploy first with 'make deploy'."; \
+		exit 1; \
+	fi
+	kubectl port-forward -n $${BUTTERCUP_NAMESPACE:-crs} service/buttercup-signoz-frontend 33301:3301 &
```
BradSwain, line 227: (nitpick) Check if the service is enabled before doing port forward

```suggestion
	fi
	@if ! kubectl get service -n $${BUTTERCUP_NAMESPACE:-crs} buttercup-signoz-frontend >/dev/null 2>&1; then \
		echo "Error: SigNoz not deployed. Set DEPLOY_SIGNOZ=true and redeploy."; \
		exit 1; \
  	fi
	kubectl port-forward -n $${BUTTERCUP_NAMESPACE:-crs} service/buttercup-signoz-frontend 33301:3301 &
```

deployment/k8s/templates/common-env.yaml
ret2libc: yes, it's a local deployment anyway. at least for now this works mainly locally. we could maybe add a warning to the README later on when we have a nicer readme

ret2libc:
- [x] Integrate with new README

michaelbrownuc:
Before we merge this we should talk about simplifying this

michaelbrownuc:
We should take this a step further before merging. Local signoz deployment should be the default for the "quick start" instructions and configuring your own external signoz should be a "manual setup" option

- [x] remove external signoz configuration questions from the `make setup-local` script
- [x] move information in the `README` about setting up external signoz to the `MANUAL SETUP` document
  - [x] We should keep information in the `README` about accessing the signoz ui via make command though
- [x] Remove the `kubectl` commands from the `README` and put them into the quick reference guide if they aren't already there. Accessing logs through the local signoz deployment should be "the way" to get logs

ret2libc:
@claude please address @michaelbrownuc 's comments

MANUAL_SETUP.md
```diff
@@ -69,6 +69,34 @@ cp deployment/env.template deployment/env
...
+3. **Alternative external observability platforms:**
+You can also configure other OTEL-compatible platforms like:
+- Jaeger
+- Zipkin
+- DataDog APM
+- New Relic
+- Honeycomb
+
+Simply set the appropriate `OTEL_ENDPOINT` and `OTEL_PROTOCOL` for your chosen platform.
+
```
ret2libc:
```suggestion
```
remove

PR #233, opened 2025-08-04
feat: Add interactive submit-project script for custom challenges

### LINKED ISSUE #220: Create a `submit-project` script/make-target to guide the user through submitting a custom challenge

ret2libc:
Implements issue #220 by adding an interactive submit-project script that guides users through submitting custom challenges.

**Features:**
- Extended challenge.py with submit-project subcommand
- Interactive questionnaire following specified requirements
- OSS-Fuzz project URL discovery using grep (no yaml import)
- Support for diff mode (..) and full mode analysis
- Automatic default branch detection
- Added Makefile target for easy access
- Validates user input and provides helpful feedback

Closes #220


orchestrator/scripts/challenge.py
```diff
@@ -548,6 +549,175 @@ def single(challenge_name: str, duration: int) -> None:
     time.sleep(duration)
 
 
+def get_project_git_url_from_oss_fuzz(oss_fuzz_url: str, oss_fuzz_ref: str, project_name: str) -> Optional[str]:
```
ret2libc, line 571: instead of using grep and sed do those things in python

ret2libc:
@claude address the review

ret2libc:
- [ ] Move this logic into buttercup-ui, so it can be triggered from the UI and we can keep everything in one place.

PR #241, opened 2025-08-04
Make release namespace and name flexible for litellm config job

reytchison:
This fixes an error in CI due to the release namespace being different.


reytchison:
Running integration test: https://github.com/trailofbits/buttercup/actions/runs/16728579979

reytchison:
The PR fixes the error during Run CRS.

PR #243, opened 2025-08-04
Scale down minikube specs and runner

reytchison:


reytchison:
Reopening because the extra `/` after `ci/` breaks the integration test. New PR: https://github.com/trailofbits/buttercup/pull/245

ret2libc:
open or not?

reytchison:
Keep closed, the new PR linked above is the one we should use.

PR #245, opened 2025-08-04
Scale down minikube specs and runner

reytchison:
Make the following changes:
- Scale down minikube requirements
- Scale down registry-cache
- Use a smaller github runner

The scaled down minikube cluster specs are based on the minimum requirements for each service.
Note that some services do not have a minimum requirement, so the cluster must be scaled to accomodate them.


reytchison:
I have not tested it on a macbook. I have only tested it on an `n2-standard-8` linux machine

PR #248, opened 2025-08-05
doc: small adjustments to the README

ret2libc:


README.md
```diff
@@ -107,10 +107,9 @@ Then navigate to `http://localhost:31323` in your web browser.
 
 In the GUI you can monitor active tasks and see when Buttercup finds bugs and generates patches for them.
 
-
 7. Stop Buttercup
 
-**Note:** This is an important step to ensure Buttercup shuts down and stops consuming third-party AI resources.
+**Note:** The system has a LLM maximum budget and tasks sent to the CRS have a fixed duration you can configure (30 minutes for the example-libpng task used above), however we still recommend you shut the system down when done with it.
```
michaelbrownuc, line 112: This is too much information I think for a "quick start" - all we really need to do here is tell them that the command is what actually turns off Buttercup. We don't really cover task durations and LLM budgets anywhere else so this level of detail is inconsistent.

PR #250, opened 2025-08-06
feat: Add local LangFuse deployment support

### LINKED ISSUE #227: Deploy a local LangFuse and log there by default
In `deployment/` we should add a LangFuse service so that when deploying the system locally LangFuse is also installed and users can see the LLM calls performed by the CRS without having access to an external LangFuse service.

The `values-upstream-minikube` template file should be changed to have something like `langfuse: deploy-enabled: true`, but it should be disabled by default. 

In `common-env` file the LANGFUSE env vars should be set to that internal LANGFUSE service by default if not specified otherwise.

ret2libc:
- Add LangFuse helm chart dependency with version 1.3.4
- Add LangFuse configuration to values.yaml with local PostgreSQL, ClickHouse, Redis, and MinIO
- Update values-upstream-minikube template to support DEPLOY_LANGFUSE flag
- Enhance common-env.yaml to automatically configure LangFuse environment variables
  - When deployed locally, uses internal service endpoints and dev credentials
  - When using external LangFuse, uses Kubernetes secrets
  - Automatically sets LANGFUSE_ENABLED=true when LangFuse is available
- Update deployment/env.template with DEPLOY_LANGFUSE configuration option

This enables users to deploy a local LangFuse instance for LLM observability without requiring external cloud services, similar to how SigNoz is integrated.

Fixes #227


ret2libc:
According to the doc, the required configuration options to set are:
```
# Optional, but highly recommended. Generate via `openssl rand -hex 32`.
#  langfuse:
#    encryptionKey:
#      value: ""
langfuse: 
  salt:
    value: secureSalt
  nextauth:
    secret:
      value: ""
 
postgresql:
  auth:
    # If you want to use `postgres` as the username, you need to provide postgresPassword instead of password.
    username: langfuse
    password: ""
 
clickhouse:
  auth:
    password: ""
 
redis:
  auth:
    password: ""
 
s3:
  auth:
    rootPassword: ""
```

deployment/k8s/values.yaml
```diff
@@ -380,3 +382,73 @@ restartPolicy: Always
...
+    enabled: true
+    auth:
+      rootPassword: "minio_dev_password"
+    resources:
+      limits:
+        cpu: 250m
+        memory: 512Mi
+      requests:
+        cpu: 100m
+        memory: 256Mi
```
ret2libc: remove this

deployment/k8s/values-upstream-minikube.template
ret2libc, line 1: add the minimum configuration for the local deployment here

deployment/k8s/templates/common-env.yaml
```diff
@@ -155,8 +155,21 @@ Define common environment for all services
...
+  value: "true"
+- name: LANGFUSE_HOST
+  value: "http://{{ .Release.Name }}-langfuse:3000"
+- name: LANGFUSE_PUBLIC_KEY
+  value: "pk-lf-local-dev-key-change-in-production"
+- name: LANGFUSE_SECRET_KEY
+  value: "sk-lf-local-dev-secret-change-in-production"
+{{- else if .Values.global.langfuse.enabled }}
+{{- /* Set Langfuse environment variables from Kubernetes secret for external service */ -}}
+- name: LANGFUSE_ENABLED
```
ret2libc, line 171: where is LANGFUSE_ENABLED used?

deployment/k8s/templates/common-env.yaml
```diff
@@ -155,8 +155,21 @@ Define common environment for all services
 Define Langfuse environment variables
 */}}
 {{- define "buttercup.env.langfuse" }}
-{{- if .Values.global.langfuse.enabled }}
-{{- /* Set Langfuse environment variables from Kubernetes secret */ -}}
+{{- if or .Values.global.langfuse.enabled .Values.global.langfuse.deployed }}
+{{- if .Values.global.langfuse.deployed }}
+{{- /* Use local deployed LangFuse service */ -}}
+- name: LANGFUSE_ENABLED
```
ret2libc, line 161: where is it used?

ret2libc:
@claude address comments from @ret2libc

deployment/k8s/values.yaml
```diff
@@ -380,3 +382,57 @@ restartPolicy: Always
...
+    auth:
+      password: "redis_dev_password"
+    master:
+      resources:
+        limits:
+          cpu: 250m
+          memory: 512Mi
+        requests:
+          cpu: 100m
+          memory: 256Mi
```
ret2libc:
```suggestion
```

PR #269, opened 2025-08-09
fix: respect OSS_FUZZ_CONTAINER_ORG environment variable

dguido:
## Summary

This PR fixes a bug in the common module where the `OSS_FUZZ_CONTAINER_ORG` environment variable was not being respected, causing test failures when using alternative container registries.

## Problem

The `oss_fuzz_container_org` property in `challenge_task.py` was hardcoded to always parse the organization from the helper_path file, ignoring the `OSS_FUZZ_CONTAINER_ORG` environment variable that should take precedence.

## Solution

Added a check to respect the `OSS_FUZZ_CONTAINER_ORG` environment variable before falling back to parsing the helper_path file:

```python
@cached_property
def oss_fuzz_container_org(self) -> str:
    # Check environment variable first
    if env_org := os.environ.get("OSS_FUZZ_CONTAINER_ORG"):
        return env_org
    
    # Fall back to parsing helper_path file
    # ... existing code ...
```

## Impact

This fix allows tests and development environments to use alternative container registries (e.g., for testing with local or private registries) by setting the `OSS_FUZZ_CONTAINER_ORG` environment variable.

## Testing

- The change has been tested locally with the environment variable set
- Existing behavior is preserved when the environment variable is not set
- Tests continue to pass in CI

## Related

See #270 for separate issue tracking the patcher integration test redesign needs.

### DIFF
--- a/deployment/crs-architecture.sh
+++ b/deployment/crs-architecture.sh
@@ -261,13 +261,18 @@ down-k8s() {
 		esac
 	fi
 	echo -e "${BLU}Deleting Kubernetes resource${NC}"
+	set -x
 	kubectl delete -k k8s/base/tailscale-connections/
 	helm uninstall --wait --namespace "$BUTTERCUP_NAMESPACE" buttercup
+	# Remove finalizers from clickhouse installation as stated in https://signoz.io/docs/operate/kubernetes/#uninstall-signoz
+	# Without this, the namespace would not be deleted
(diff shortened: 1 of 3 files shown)


ret2libc (approved):
We don't have `OSS_FUZZ_CONTAINER_ORG` described anywhere at the moment, I think, but I think this is still nice to have as it allows to force a specific container if necessary.

PR #286, opened 2025-08-14
Fix button width at all times.

evandowning:
When refreshing the webpage, the button widths/heights for `Submit Task`, `Submit libpng task`, and `Refresh` all change sizes.

This is visually distracting to users.

This PR fixes the button sizes dynamically, based on the page's first load.


michaelbrownuc (approved):
LGTM

PR #292, opened 2025-08-18
ci: make linting truly conditional

ret2libc:


ret2libc:
I've tested this with no changes and with changes only to the patcher and it worked in both cases.

PR #293, opened 2025-08-18
fix: restrict Claude PR review to labeled PRs only

dguido:
## Summary
- Modified the Claude code review workflow to only run when PRs are explicitly labeled with `claude-review`
- This reduces noise from automatic reviews on every PR

## Changes
- Added `labeled` event type to the workflow triggers
- Added conditional check: `if: contains(github.event.pull_request.labels.*.name, 'claude-review')`
- Workflow now only runs when the `claude-review` label is present on the PR

## How to use
1. Create or update a PR
2. Add the `claude-review` label when you want Claude to review it
3. The review will automatically trigger

This change makes the Claude review bot opt-in rather than automatic, reducing notification noise while still allowing teams to request AI reviews when needed.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -2,7 +2,7 @@ name: Claude Code Review
 
 on:
   pull_request:
-    types: [opened, synchronize, labeled]
+    types: [labeled, synchronize]
     # Optional: Only run on specific file changes
     # paths:
     #   - "src/**/*.ts"


.github/workflows/claude-code-review.yml
```diff
@@ -2,7 +2,7 @@ name: Claude Code Review
 
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [opened, synchronize, labeled]
```
michaelbrownuc: nit: the conditional will still prevent this from running but we don't need to fire the workflow on every PR open if it won't do anything

```suggestion
    types: [labeled, synchronize]
```

PR #306, opened 2025-08-20
Support Gemini

Boyan-MILANOV:
This PR adds support for google Gemini models.

- Includes: `gemini-pro`, `gemini-2.5-flash`, `gemini-2.5-flash-exp`
- Users can now enter a `GEMINI_API_KEY` secret during setup
- All components now have `gemini-pro` added to their list of fallback models

Tested by following the`README` instructions: doing a fresh install, adding only the gemini API key, running the example task, verifying that a successful PoV + patch where created using the buttercup UI.

### DIFF
--- a/common/src/buttercup/common/llm.py
+++ b/common/src/buttercup/common/llm.py
@@ -5,8 +5,8 @@ from enum import Enum
 from typing import Any
 
 import requests
-from langchain_core.language_models import BaseChatModel
 from langchain.callbacks.base import BaseCallbackHandler
+from langchain_core.language_models import BaseChatModel
 from langchain_core.runnables import ConfigurableField, Runnable
 from langchain_openai.chat_models import ChatOpenAI
 from langfuse.callback import CallbackHandler


common/src/buttercup/common/llm.py
```diff
@@ -32,6 +32,9 @@ class ButtercupLLM(Enum):
     CLAUDE_3_5_SONNET = "claude-3.5-sonnet"
     CLAUDE_3_7_SONNET = "claude-3.7-sonnet"
     CLAUDE_4_SONNET = "claude-4-sonnet"
+    GEMINI_PRO = "gemini-pro"
+    GEMINI_2_FLASH = "gemini-2.0-flash"
+    GEMINI_2_FLASH_EXP = "gemini-2.0-flash-exp"
```
ret2libc: isn't there 2.5 already?

patcher/src/buttercup/patcher/agents/context_retriever.py
```diff
@@ -583,6 +583,7 @@ def sh(
 
 def _are_test_instructions_valid(instructions: str, output: bytes, error: bytes) -> bool:
     """Validate a set of test instructions by executing them inside the project environment."""
+    # TODO(boyan): default model
```
ret2libc: what does this mean?

common/src/buttercup/common/llm.py
Boyan-MILANOV: I based myself on the litellm docs and they don't have it yet https://docs.litellm.ai/docs/providers/gemini#chat-models

deployment/k8s/values.yaml
```diff
@@ -375,6 +375,24 @@ litellm-helm:
...
+      - model_name: gemini-2.0-flash-exp
+        litellm_params:
+          model: gemini/gemini-2.0-flash-exp
+          api_key: os.environ/GEMINI_API_KEY
+
+      - model_name: gemini-2.0-flash
+        litellm_params:
+          model: gemini/gemini-2.0-flash
+          api_key: os.environ/GEMINI_API_KEY
+        # tpm and rpm??
```
Boyan-MILANOV: @ret2libc should we set tpm and rpm values for Gemini? How where values for other models chosen?

patcher/src/buttercup/patcher/agents/common.py
```diff
@@ -426,6 +426,7 @@ def __eq__(self, other: object) -> bool:
 def _create_understand_code_snippet_chain() -> Runnable:
     return (  # type: ignore[no-any-return]
         UNDERSTAND_CODE_SNIPPET_PROMPT
+        # TODO(boyan): set this to gemini if openAI not available?
         | create_default_llm_with_temperature(model_name=ButtercupLLM.OPENAI_GPT_4_1.value)
```
Boyan-MILANOV: @ret2libc should we add fallbacks to this? Or directly to the `create_default_llm_with_temperature` function perhaps?

deployment/k8s/values.yaml
michaelbrownuc: The values were chosen based on our allowable tiers, which were maxed out.  Please use paid tier rate limits locted here: https://ai.google.dev/gemini-api/docs/pricing

common/src/buttercup/common/llm.py
ret2libc: They also say "We support ALL Gemini models, just set model=gemini/<any-model-on-gemini> as a prefix when sending litellm requests" so probably they do support it

patcher/src/buttercup/patcher/agents/common.py
ret2libc: yes please add a fallback to this. we should fallback to a claude and then to a gemini model. Also for other places where we only use one llm provider

Boyan-MILANOV: Done.

patcher/src/buttercup/patcher/agents/context_retriever.py
Boyan-MILANOV: Residual comment indicating I should add fallback models there. Resolved.

common/src/buttercup/common/llm.py
Boyan-MILANOV: Updated to 2.5 for flash models. For gemini pro it doesn't matter as the version isn't specified so the latest gets selected automatically

michaelbrownuc:
@Boyan-MILANOV the full system integration tests are consistently failing at the patching step, but are passing on main. Can you investigate?

Boyan-MILANOV:
Yep I looking at it and trying to reproduce locally.

PR #308, opened 2025-08-20
fix: remove insecure URL substring check for GitHub authentication

michaelbrownuc:
Remove the redundant and insecure 'github.com' in repo_url check that could match malicious URLs. The inner startswith('https://github.com/') check already provides proper security validation.

This addresses CodeQL security warning py/incomplete-url-substring-sanitization.

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
+++ b/orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
@@ -72,7 +72,7 @@ class ChallengeService:
             github_pat = os.environ.get("GITHUB_PAT")
             github_username = os.environ.get("GITHUB_USERNAME")
 
-            if github_pat and github_username and "github.com" in repo_url:
+            if github_pat and github_username:
                 # For GitHub repositories, use the PAT for authentication
                 if repo_url.startswith("https://github.com/"):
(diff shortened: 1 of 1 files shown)


orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
```diff
@@ -72,7 +72,7 @@ def create_challenge_tarball(
             github_pat = os.environ.get("GITHUB_PAT")
             github_username = os.environ.get("GITHUB_USERNAME")
 
-            if github_pat and github_username and "github.com" in repo_url:
+            if github_pat and github_username:
                 # For GitHub repositories, use the PAT for authentication
                 if repo_url.startswith("https://github.com/"):
```
hbrodin: There are only two cases now, either `gh_pat and gh_username and repo_url.starts...`, or not. Could simplify the if-else structure.

michaelbrownuc: yeah, this is the original suggestion by claude, I think we should move the `repo_url.startswith` condition into the top level `if` statement and simplify the rest. This would be most clear.

michaelbrownuc: @claude - fix this snippet to move the condition `repo_url.startswith("https://github.com/")` into the top level conditional statement and simplify the rest of the `else` statements

PR #295, opened 2025-08-18
Fix ruff warnings and add pre-commit hooks

dguido:
## Summary
This PR standardizes Python project configuration and improves code quality across the codebase.

### 📦 Standardized Python Project Configuration
- Aligned all `pyproject.toml` files with consistent structure and metadata
- Standardized ruff configuration across all components (line-length, target-version, lint rules)
- Enabled stricter ruff rules that the codebase now passes: UP009, UP012, UP015, UP032
- Added targeted ignore rules for legacy clusterfuzz code

### 🔧 Pre-commit Infrastructure
- Added comprehensive `.pre-commit-config.yaml` with standard hooks
- Fixed pre-commit CI workflow by installing `uv` in GitHub Actions
- Updated pre-commit-hooks from v4.6.0 to v6.0.0 to resolve deprecated warnings
- Added Python 3.12 compatibility in CI workflow

### ✨ Code Quality Improvements
- Applied all safe ruff auto-fixes across the entire codebase
- Fixed import sorting and formatting issues
- Removed redundant code patterns (RET501, RET505, PIE790, etc.)
- Applied code simplification rules (B905, C416, SIM103, SIM114)
- Updated `.gitignore` to exclude Python cache directories

### 🔒 Security Fix
- Removed insecure URL substring check for GitHub authentication (CodeQL warning)

### 🛠️ Developer Experience
- Enhanced `scripts/lint-changed-files.sh` with JSON output for AI tools
- Added comprehensive test suite for lint script validation
- Documented AI tool integration patterns

## Result
All components now have:
- Consistent project metadata and dependency management
- Standardized linting and formatting configuration
- Working pre-commit hooks in CI
- Cleaner, more maintainable code
- Resolved security warnings

PR #312, opened 2025-08-20
chore: add mypy cache to gitignore

dguido:
## Summary
This PR adds mypy cache directories to .gitignore to keep the repository clean.

## Changes
- Added `.mypy_cache/` pattern to .gitignore
- Added `**/.mypy_cache/` to catch nested cache directories

## Rationale
Mypy creates cache directories for type checking that should not be committed:
- They are environment-specific
- They can cause conflicts between different Python versions
- They add unnecessary noise to the repository

This is a small housekeeping change extracted from PR #295.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,10 @@ dist/
 pdm.lock
 .langchain.db
 
+# Mypy type checking cache
+.mypy_cache/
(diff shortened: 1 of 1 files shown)


michaelbrownuc (approved):
LGTM

PR #320, opened 2025-08-22
fuzzer_runner: split fuzzer's dependencies

ret2libc:
Before this commit, the fuzzer dependend on clusterfuzz, which uses
protobuf 3.20. Since fuzzer-bot depended on common subpackage as well
and also common (and the other packages) require protobuf, we had to use
protobuf 3.20 everywhere. This old dependency however means that a lot
of packages can't be used in their "newer" versions, because they depend
on newer protobuf versions.

This commit splits the fuzzer into a separate fuzzer-runner that is
executed in a separate process inside a separate venv. fuzzer-runner
executes the clusterfuzz-heavy operations (e.g. fuzzing) and isolates the
clusterfuzz dependency, so that the rest of the system can use newer
protobuf version.

* split `fuzzer-bot` in `fuzzer-runner` and `fuzzer-bot`
* have a "full" optional dependency group in `common`, including `openlit`
  and `protobuf`, so that `fuzzer-runner` can use the lite version of `common`
  without bringing those heavy deps
* move `FuzzConfiguration`/`BuildConfiguration` in a separate common file
  that doesn't require to load all protobuf files. Again, in this way
  other components can just depend on the `common` "lite" version and not
  require protobuf stuff
* add `fuzzer-runner` as a separate venv inside the `fuzzer-bot` container
* add `RunnerProxy` class in fuzzer package to provide an interface to
  interact with the fuzzer-runner binary.

---

Reasons for this:
- we couldn't update to newer openlit because it required newer protobuf
- we had some hack to make openlit work, and we can now remove them
- i wanted to test other easier alternatives to langchain for the patcher agents (e.g. pydantic-ai), but they require newer deps that we can't get because of protobuf==3.20


fuzzer_runner/src/buttercup/fuzzer_runner/server.py
```diff
@@ -0,0 +1,235 @@
...
+        "timeout": request.timeout or (server_settings.timeout if server_settings else 1000),
+    }
+
+    # Run fuzzer in background
+    background_tasks.add_task(_run_fuzzer_task, task_id, fuzz_conf, request.timeout)
+
+    return FuzzResponse(task_id=task_id, status="running")
+
+
+@app.post("/merge-corpus", response_model=MergeCorpusResponse)
```
hbrodin: Is there any way of making the paths into constants shared by the proxy and runner?

fuzzer_runner/README.md
```diff
@@ -0,0 +1,177 @@
+# Fuzzer Runner
```
hbrodin, line 1: I haven't thought too deeply about this, but would this be a reusable component if we were to introduce additional fuzzers such as AFL or similar? Or would this have to change again then?

ret2libc: I think those fuzzers are already supported in clusterfuzz and they return the same objects as libfuzzer, don't they? If they do, it should be straightforward to extend it. If they don't, we might need to change a bit the interfaces.

common/src/buttercup/common/logger.py
```diff
@@ -1,19 +1,24 @@
 import os
 
-from opentelemetry._logs import set_logger_provider
+try:
```
hbrodin, line 3: This change appears kind of unrelated? Or?

common/src/buttercup/common/telemetry.py
```diff
@@ -4,28 +4,22 @@
...
-
-
-def _clean_attribute_wrapper(key: str, value: Any, max_len: int | None = None) -> Any:
-    """Wrapper around _clean_attribute to add custom behavior"""
-    if isinstance(value, ChatPromptValue):
-        value = value.to_string()
+logger = logging.getLogger(__name__)
 
-    return _clean_attribute_orig(key, value, max_len)
+try:
```
hbrodin, line 9: Same here (unrelated). Are these part of a separate PR?

common/src/buttercup/common/logger.py
ret2libc: It is not, it's related to what's described in the first comment:

> have a "full" optional dependency group in common, including openlit
and protobuf, so that fuzzer-runner can use the lite version of common
without bringing those heavy deps

IOW with these changes we can still use some parts of `common` also in `fuzzer-runner` without requiring opentelemetry and protobuf old version.

deployment/k8s/templates/common-env.yaml
```diff
@@ -338,6 +338,13 @@ Service-specific environment variables that utilize the standardized variables
   value: {{ int .Values.global.maxPovSize | quote }}
 {{- end }}
 
+{{- define "buttercup.fuzzerRunnerEnv" }}
+- name: BUTTERCUP_FUZZER_RUNNER_HOST
+  value: "0.0.0.0"
```
hbrodin: I forget exactly how it works with the sidecar containers now, but assuming we end up with two fuzzers on the same node, will the runner-proxy be able to figure out which fuzzer-runner instance to reach? I assume it works implicitly on a single node-single fuzzer instance deployment.

ret2libc: I will double check this, good point 👍 Ideally the container just sees the sidecar container connection (maybe we have to switch to `127.0.0.1`).

fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
```diff
@@ -0,0 +1,184 @@
...
+class Conf:
+    # in seconds
+    timeout: int
+    server_url: str = "http://localhost:8000"
+    poll_interval: float = 1.0  # seconds between status checks
+    timeout_buffer: float = 60.0  # extra time buffer beyond the fuzzer timeout
+
+
+@dataclass
+class RunnerProxy:
```
hbrodin, line 43: How does this deal with a blocked runner? IIRC we had a few cases where the runner could get stuck. I believe we previously used a health-script for that and just restarted the pod.

fuzzer_runner/src/buttercup/fuzzer_runner/server.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+@app.get("/health", response_model=HealthResponse)
+async def health_check() -> HealthResponse:
+    """Health check endpoint"""
+    return HealthResponse(status="healthy", version="0.1.0")
+
+
+@app.post("/fuzz", response_model=FuzzResponse)
+async def run_fuzzer(request: FuzzRequest, background_tasks: BackgroundTasks) -> FuzzResponse:
```
hbrodin: Consider implementing a helper function as most of the lines of run_fuzzer and merge_corpus are identical

fuzzer_runner/src/buttercup/fuzzer_runner/server.py
```diff
@@ -0,0 +1,235 @@
...
+    return {
+        "task_id": task_id,
+        "type": task_info["type"],
+        "status": task_info["status"],
+        "result": task_info.get("result"),
+        "error": task_info.get("error"),
+    }
+
+
+@app.get("/tasks", response_model=dict[str, Any])
```
hbrodin: Should there be an operation to `kill` a task? E.g. for process health monitoring. I kind of assume that there will always be just one task running at a time? Or?

patcher/src/buttercup/patcher/agents/leader.py
```diff
@@ -36,7 +36,7 @@ class PatcherLeaderAgent(PatcherAgentBase):
     tasks_storage: Path
     model_name: str | None = None
 
-    def _init_patch_team(self) -> StateGraph[PatcherAgentState, PatcherAgentState, PatcherAgentState]:
+    def _init_patch_team(self) -> StateGraph[PatcherAgentState, PatcherConfig, PatcherAgentState, PatcherAgentState]:
```
hbrodin, line 39: Was PatcherConfig just missing before?

seed-gen/src/buttercup/seed_gen/vuln_base_task.py
```diff
@@ -256,13 +256,12 @@ def submit_valid_pov(
...
-            crash_input_path=dst,
-            stacktrace=stacktrace,
-            crash_token=ctoken,
-        )
+        crash = Crash()
+        crash.target.CopyFrom(build)
+        crash.harness_name = self.harness_name
+        crash.crash_input_path = dst
+        crash.stacktrace = stacktrace
+        crash.crash_token = ctoken
```
hbrodin, line 264: Why the change?

common/src/buttercup/common/logger.py
hbrodin: Ahh. Makes sense. Thanks!

patcher/src/buttercup/patcher/agents/leader.py
ret2libc: no, I think it's due to new langgraph version

seed-gen/src/buttercup/seed_gen/vuln_base_task.py
ret2libc: It seems to be related to the new protobuf version :/ I couldn't assign things like before anymore as it was complaining at runtime.

fuzzer_runner/src/buttercup/fuzzer_runner/server.py
ret2libc: Unfortunately there isn't a way to kill a `BackgroundTask` (as implemented right now). However, I want to change the `asyncio.to_thread` where I call the actual clusterfuzz function and have a:
```
            result: FuzzResult = await asyncio.wait_for(
                asyncio.to_thread(runner.run_fuzzer, fuzz_conf),
                timeout=future_timeout
            )
```
which should just raise a `asyncio.TimeoutError` when the fuzzer is stuck. That means essentially if things get stuck in the fuzzer we just wait for the timeout to finish and at that point the future will be stopped.

fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
ret2libc: There is the timeout enforced in `_wait_for_task_completion` and the runner also has (will have soon) a future that can wait for at most a given timeout, so if things get stuck they should be stopped anyway after the timeouts expire.

deployment/k8s/templates/common-env.yaml
ret2libc: switched to 127.0.0.1 and each container can only see its side container.

fuzzer_runner/src/buttercup/fuzzer_runner/server.py
hbrodin: Sounds like a good path forward. There should really be a limited number of threads so I don't seen that being an issue either.

fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
hbrodin: Sounds good. Given the current implementation, we would then just start a new fuzzer instance in parallel to the old one that is stuck (as there isn't any way of killing it at the moment)?

ret2libc:
Thanks for the review @hbrodin ! I'll keep this as a draft for now to avoid merging it, I want to test it a bit more and make sure your concerns are addressed. I think they should, as the asyncio thread should be canceled after the timeout triggers, but I need to double check.

fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
```diff
@@ -0,0 +1,215 @@
...
+        try:
+            process = subprocess.Popen(
+                cmd,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                start_new_session=True,
+            )
+
+            try:
+                stdout, stderr = process.communicate(timeout=subprocess_timeout)
```
hbrodin, line 93: Is there a risk that either stdout or stderr becomes huge here? Should we consider a limit on the amount of data read? IIRC `.communicate()` is how you are supposed to do things, but just thinking that we might want to cut in the beginning or middle of the output if it becomes huge and that would probably require explicit reading of the fds. Maybe create an issue and ignore for now?
Maybe such an issue would manifest at earlier stages first, in the runner?

dev/docker-compose/env.dev.compose
```diff
@@ -46,10 +46,15 @@ BUTTERCUP_PROGRAM_MODEL_LOG_LEVEL=debug
 BUTTERCUP_PROGRAM_MODEL_SCRATCH_DIR=/node_data/crs_scratch
 BUTTERCUP_PROGRAM_MODEL_GRAPHDB_ENABLED=false
 
+BUTTERCUP_FUZZER_REDIS_URL=redis://redis:6379
```
hbrodin, line 49: Why are these needed now and not earlier? Or is it just some convenience thing?

hbrodin (approved):
Looks good! Well done! One thing to consider before merging is the handling of stdout and stderr, but it is not critical to fix now I think. Might even need to be fixed. But wanted to at least bring it up.

dev/docker-compose/env.dev.compose
ret2libc: those settings were passed as CLI arguments before, i removed those and set them here

fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
ret2libc: Yes, good point. I can fix it quickly I think.

ret2libc: Created #350

PR #326, opened 2025-08-27
Parse result logs

evandowning:
This adds scripts and instructions on how to parse logs from the final round of the competition.

- [x] Write scripts to unpack results files/logs
- [x] Write script to parse PoVs and Patches
- [x] Reorg scripts to the root `scripts` directory


michaelbrownuc:
We should probably add the scripts to a different directory. These scripts really aren't related to the orchestrator. They probably belong in the root scripts directory. If that directory is getting too crowded, we should organize with subfolders.

Also - if there are any other scripts unrelated to actual orchestrator functions in the `orchestrator/scripts` directory, please move those as well.

michaelbrownuc (approved):
LGTM as is, but we need further re-org before merging. Please add a root level folder called `guides` and move the various informative `.md` files (except readme, contributing, and claude) there. We have too many markdown files in the root rn.

evandowning:
> LGTM as is, but we need further re-org before merging. Please add a root level folder called `guides` and move the various informative `.md` files (except readme, contributing, and claude) there. We have too many markdown files in the root rn.

Done. Let me know what you think.

michaelbrownuc (approved):
LGTM!

PR #330, opened 2025-09-03
Allows the user to accept/reject the generated patch

### LINKED ISSUE #329: Allow user to accept or reject patch
It would be helpful as a user to be able to reject a generated patch in case it does not behave/look as desired. The functionality is already there on the CRS side, as the compeitition-api had the functionality. This would only require changing the UI to not automatically approve the patch, but rather have the user do it.

While a patch is not approved there is no additional work done. When the patch is rejected, a request for a new patch is generated and eventually a new patch is submitted that can be accepted/rejected.

Longer term, one could envision providing feedback to the patcher component. This is a first step.

hbrodin:
Closes #329.

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/ui/database.py
+++ b/orchestrator/src/buttercup/orchestrator/ui/database.py
@@ -81,6 +81,7 @@ class Patch(Base):
     patch_id: Mapped[str] = mapped_column(String, primary_key=True, default=lambda: str(uuid.uuid4()))
     task_id: Mapped[str] = mapped_column(String, ForeignKey("tasks.task_id"), nullable=False)
     patch: Mapped[str] = mapped_column(Text)
+    status: Mapped[str] = mapped_column(String, default="accepted")  # accepted, passed, failed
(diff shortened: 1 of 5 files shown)


ret2libc:
Same request as for the other PR, any image to show the new behaviour?

hbrodin:
Screenshot 2025-09-04 at 09 46 31
Patches view
Screenshot 2025-09-04 at 09 46 41
After selecting patch
Screenshot 2025-09-04 at 09 46 54
After clicking "Reject"

orchestrator/src/buttercup/orchestrator/ui/database.py
```diff
@@ -118,6 +119,24 @@ def _create_tables(self) -> None:
         Base.metadata.create_all(bind=self.engine)
         logger.info("Database tables created/verified")
 
+        # Migrate existing patches to have status field
+        self._migrate_patch_status()
+
+    def _migrate_patch_status(self) -> None:
```
ret2libc, line 125: hehe i realized this problem, but I didn't want to introduce something like Alembic at the time. Maybe we should, at some point, if we want to go further with this UI

PR #333, opened 2025-09-03
Move the example-libpng button

### LINKED ISSUE #331: UI simplification
Remove the `Submit example-libpng task` button and instead in the `Submit Task dialog` add a button that fills in those values so you can modify e.g. timeout easily and only keeping one flow.

hbrodin:
There is now an additional button available in the Submit Task dialogue. This enables users to easily modify values such as timeouts etc and keeps the main ui clean.
Closes #331


ret2libc:
Could you provide images of the final result when doing modifications to UI?

hbrodin:
Screenshot 2025-09-03 at 15 22 34
Screenshot 2025-09-03 at 15 22 23
Screenshot 2025-09-03 at 15 22 14

ret2libc (approved):

PR #335, opened 2025-09-03
ci: do not use ghcr auth

ret2libc:


ret2libc:
CI on main was failing because of the PAT I used expired, but apparently we don't need it anymore because https://github.com/tob-challenges/example-libpng is public.

.github/workflows/integration.yml
```diff
@@ -91,6 +91,7 @@ jobs:
         run: |
           cp ../.github/ci-env.template env
 
+          sed -i "/GHCR_AUTH=.*/d" env
```
hbrodin, line 94: Are we using `GHCR_AUTH` anywhere else now? Or should it be removed from the template files as well?

ret2libc: it is still useful if you need to access private repos. I wouldn't remove it for now.

PR #336, opened 2025-09-03
Test and improve AKS deployment

ret2libc:


deployment/backend.tf
```diff
@@ -1,7 +1,7 @@
 terraform {
   backend "azurerm" {
-    resource_group_name  = "example-tfstate-rg"
-    storage_account_name = "tffinalacc"
+    resource_group_name  = "aixcc"
+    storage_account_name = "buttercuptf42cv"
     container_name       = "tfstate"
     key                  = "terraform.tfstate"
```
ret2libc: revert

Makefile
```diff
@@ -96,8 +84,8 @@ crs-instance-id:
 	fi
 	echo "CRS instance ID: $$(kubectl get configmap -n $${BUTTERCUP_NAMESPACE:-crs} crs-instance-id -o jsonpath='{.data.crs-instance-id}')"
 
-deploy-local:
-	@echo "Deploying to local Minikube environment..."
+deploy:
+	@echo "Deploying environment..."
```
hbrodin, line 88: Line 90 should probably be modified to say something other than Run 'make setup-local' first.'

Makefile
```diff
@@ -106,25 +94,19 @@ deploy-local:
...
+	@read ans; \
+	ans_lc=$$(echo "$$ans" | tr '[:upper:]' '[:lower:]'); \
+	if [ "$$ans_lc" != "yes" ] && [ "$$ans_lc" != "y" ]; then \
+		echo "Aborted by user."; \
 		exit 1; \
 	fi
 	cd deployment && make up
-	crs_instance_id=$$(make crs-instance-id)
-	echo "CRS instance ID: $$crs_instance_id"
+	make crs-instance-id
```
hbrodin, line 108: Will this already print it?

README.md
```diff
@@ -23,7 +23,7 @@
 
 **Note:** Buttercup uses third-party AI providers (LLMs from companies like OpenAI, Anthropic and Google), which cost money. Please ensure that you manage per-deployment costs by using the built-in LLM budget setting.
 
-**Note:** Buttercup works best with access to models from OpenAI **and** Anthropic, but can be run with at least one API key from one third-party provider (support for Gemini coming soon).
+**Note:** Buttercup works best with access to models from OpenAI **and** Anthropic, but can be run with at least one API key from one third-party provider. You can a combination of OpenAI, Anthropic, and Google LLMs.
```
hbrodin:
```suggestion
**Note:** Buttercup works best with access to models from OpenAI **and** Anthropic, but can be run with at least one API key from one third-party provider. You can use a combination of OpenAI, Anthropic, and Google LLMs.
```

scripts/setup-azure.sh
```diff
@@ -23,44 +23,92 @@ setup_service_principal() {
...
-        print_success "Service principal created successfully"
-        echo
-        print_status "Service Principal Details:"
-        echo "  Name: $sp_name"
-        echo "  App ID: $app_id"
-        echo "  Tenant ID: $tenant_id"
-        echo "  Password: $password"
+
+    # Source deployment/env to get existing RESOURCE_GROUP_NAME if set
+    if [ -f "deployment/env" ]; then
```
hbrodin, line 28: Should we output a warning and ask the user if they want to create one before starting to do the individual steps?

scripts/setup-azure.sh
```diff
@@ -23,44 +23,92 @@ setup_service_principal() {
...
+        read -p "Do you want to create it? (Y/n): " -n 1 -r
         echo
-        print_warning "Save these credentials securely!"
+        if [[ $REPLY =~ ^[Nn]$ ]]; then
+            print_error "Resource group must exist before proceeding"
+            exit 1
+        fi
+
+        print_status "Creating resource group: $RESOURCE_GROUP_NAME"
+        az group create --name "$RESOURCE_GROUP_NAME" --location eastus
```
hbrodin: Maybe we should make location/region a variable as well?

hbrodin (approved):
I have done a first pass reading it through but overall it looks good. Haven't tried to run it. I believe we can move on with this one. I left a few comments with minor suggestions/questions.

Makefile
ret2libc: yes!

Makefile
```diff
@@ -96,8 +84,8 @@ crs-instance-id:
 	fi
 	echo "CRS instance ID: $$(kubectl get configmap -n $${BUTTERCUP_NAMESPACE:-crs} crs-instance-id -o jsonpath='{.data.crs-instance-id}')"
 
-deploy-local:
-	@echo "Deploying to local Minikube environment..."
+deploy:
+	@echo "Deploying environment..."
 	@if [ ! -f deployment/env ]; then \
 		echo "Error: Configuration file not found. Run 'make setup-local' first."; \
```
ret2libc:
```suggestion
		echo "Error: Configuration file not found. Run 'make setup-local' or `make setup-azure` first."; \
```

scripts/setup-azure.sh
ret2libc: Might be better, but this function is called only after `setup_config`, thus the config file should be there already.

PR #341, opened 2025-09-04
patcher: fix run_find_tests_agent timeout

ret2libc:
For some reasons, while testing the patcher locally the timeout was not respected (even though I think it was during the competition).

Relying on asyncio fixes the problem.
Plus, this makes the find-tests timeout configurable by setting an `TOB_PATCHER....` env var.

### DIFF
--- a/patcher/src/buttercup/patcher/agents/config.py
+++ b/patcher/src/buttercup/patcher/agents/config.py
@@ -18,6 +18,7 @@ class PatcherConfig(BaseModel):
     max_root_cause_analysis_retries: int = Field(default=3)
     max_patch_strategy_retries: int = Field(default=3)
     max_tests_retries: int = Field(default=5)
+    find_tests_timeout_min: int = Field(default=30)
     ctx_retriever_recursion_limit: int = Field(default=80)
     patch_validation_recursion_limit: int = Field(default=30)
     n_initial_stackframes: int = Field(default=4)
--- a/patcher/src/buttercup/patcher/agents/context_retriever.py
+++ b/patcher/src/buttercup/patcher/agents/context_retriever.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import asyncio
 import operator
 import re
 import tempfile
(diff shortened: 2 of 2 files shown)


patcher/src/buttercup/patcher/agents/config.py
```diff
@@ -18,6 +18,7 @@ class PatcherConfig(BaseModel):
     max_root_cause_analysis_retries: int = Field(default=3)
     max_patch_strategy_retries: int = Field(default=3)
     max_tests_retries: int = Field(default=5)
+    find_tests_timeout_min: int = Field(default=30)
```
hbrodin: I guess `.._min` should be interpreted as minutes. However it is also a common abbreviation for minimum. I think we are using seconds in most other places with timeouts, or? Not important, just a thought.

patcher/src/buttercup/patcher/agents/context_retriever.py
```diff
@@ -1299,46 +1300,43 @@ def run_find_tests_agent() -> FindTestsState | None:
...
-                        )
-                        self._save_custom_test_instructions(agent_state.tests_instructions)
-
-                        return Command(
-                            update={
-                                "tests_instructions": agent_state.tests_instructions,
-                            },
-                            goto=PatcherAgentName.ROOT_CAUSE_ANALYSIS.value,
-                        )
+                agent_state = asyncio.run(asyncio.wait_for(run_find_tests_agent(), timeout=timeout_seconds))
```
hbrodin: Caveat: I haven't looked into the details of `find_test_agent` recently. But, is there a risk that when it runs, it gets blocked on something (that doesn't play nice with asyncio), the timeout occurs here, this code continues and somehow gets back to modifying `find_test_agent` again while it potentially runs or is resumed elsewhere? I.e. a data race (-ish).

hbrodin (approved):
Looks good! I left a couple of comments for your consideration.

patcher/src/buttercup/patcher/agents/config.py
ret2libc: makes sense

ret2libc:
It seems langgraph doesn't play well with this...

PR #353, opened 2025-09-12
Generate random litellm/crs keys

### LINKED ISSUE #351: Generate CRS_KEY_TOKEN/CRS_KEY_TOKEN_HASH during the setup scripts

ret2libc:
Closes #351

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -549,7 +549,7 @@ configure_local_api_keys() {
 }
 
 generate_litellm_master_key() {
-    # If CRS_KEY_ID is already set to the default value, skip key generation
+    # If LITELLM_MASTER_KEY is already configured, skip key generation
     if [ -f "deployment/env" ]; then
         source deployment/env
     fi
@@ -566,7 +566,7 @@ generate_litellm_master_key() {
 }
 
 generate_crs_key_id_token() {
-    # If CRS_KEY_ID is already set to the default value, skip key generation
+    # If CRS_KEY_ID is already configured, skip key generation
     if [ -f "deployment/env" ]; then
         source deployment/env
     fi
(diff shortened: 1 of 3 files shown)


scripts/common.sh
```diff
@@ -548,6 +548,46 @@ configure_local_api_keys() {
     print_success "API keys configured successfully"
 }
 
+generate_litellm_master_key() {
+    # If CRS_KEY_ID is already set to the default value, skip key generation
```
hbrodin: This comment looks a bit out of place. I assume we are talking about `LITELLM_MASTER_KEY` here and that we proceed with key-generation if it is empty or set to default key.

scripts/common.sh
```diff
@@ -548,6 +548,46 @@ configure_local_api_keys() {
...
+
+    print_linebreak
+    print_status "Configuring LiteLLM master key..."
+    local litellm_master_key=$(openssl rand -hex 16)
+    portable_sed "s|.*export LITELLM_MASTER_KEY=.*|export LITELLM_MASTER_KEY=\"$litellm_master_key\"|" deployment/env
+    print_success "LiteLLM master key configured successfully"
+}
+
+generate_crs_key_id_token() {
+    # If CRS_KEY_ID is already set to the default value, skip key generation
```
hbrodin: It appears as we skip key generation if the `CRS_KEY_ID` is _not_ set to default value.

hbrodin (approved):
Looks good! Maybe just update the comments a bit to reflect what happens in the code :)

scripts/common.sh
ret2libc:
```suggestion
    # If LITELLM_MASTER_KEY is already configured, skip key generation
```

scripts/common.sh
ret2libc:
```suggestion
    # If CRS_KEY_ID is already configured, skip key generation
```

PR #354, opened 2025-09-12
Add Tailscale support to the custom scripts

### LINKED ISSUE #344: Configure Tailscale on AKS deployment

ret2libc:
Closes #344

### DIFF
--- a/scripts/web_ui.sh
+++ b/scripts/web_ui.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+if [ ! -f deployment/env ]; then
+    echo "Error: Deployment environment file not found. Run 'make setup-local' or 'make setup-azure' first."
+    exit 1
+fi
+source ./deployment/env
+
+if ! kubectl get namespace "${BUTTERCUP_NAMESPACE:-crs}" >/dev/null 2>&1; then
+    echo "Error: CRS namespace not found. Deploy first with 'make deploy'."
+    exit 1
+fi
+
+echo "Opening web UI..."
+
+if [ "${TAILSCALE_ENABLED:-false}" = "true" ]; then
+    # Use Tailscale domain
+    if [ -z "${TAILSCALE_DOMAIN}" ]; then
+        echo "Error: TAILSCALE_ENABLED is true but TAILSCALE_DOMAIN is not set"
+        exit 1
+    fi
+    CRS_HOSTNAME="${CRS_HOSTNAME:-buttercup-ui}"
+    UI_URL="http://${CRS_HOSTNAME}.${TAILSCALE_DOMAIN}"
+else
+    # Use local port-forward
+    kubectl port-forward -n "${BUTTERCUP_NAMESPACE:-crs}" service/buttercup-ui 31323:1323
+    sleep 3
+    UI_URL="http://localhost:31323"
+fi
+
+echo "Opening web UI at $UI_URL..."
+if command -v xdg-open >/dev/null 2>&1; then
+    xdg-open "$UI_URL"
+elif command -v open >/dev/null 2>&1; then
+    open "$UI_URL"
+else
+    echo "Please open $UI_URL in your browser."
+fi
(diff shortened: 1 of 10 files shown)


scripts/web_ui.sh
```diff
@@ -0,0 +1,38 @@
...
+    # Use Tailscale domain
+    if [ -z "${TAILSCALE_DOMAIN}" ]; then
+        echo "Error: TAILSCALE_ENABLED is true but TAILSCALE_DOMAIN is not set"
+        exit 1
+    fi
+    CRS_HOSTNAME="${CRS_HOSTNAME:-buttercup-ui}"
+    UI_URL="http://${CRS_HOSTNAME}.${TAILSCALE_DOMAIN}"
+else
+    # Use local port-forward
+    kubectl port-forward -n "${BUTTERCUP_NAMESPACE:-crs}" service/buttercup-ui 31323:1323
```
hbrodin: Isn't this running in the foreground and will block further progress?

hbrodin (commented):
Overall, looks good! Good move fixing this! Just one comment I'd like to clear before moving on.

scripts/web_ui.sh
ret2libc: oh nice catch :D

ret2libc:
```suggestion
    kubectl port-forward -n "${BUTTERCUP_NAMESPACE:-crs}" service/buttercup-ui 31323:1323 &
```

scripts/web_ui.sh
```diff
@@ -0,0 +1,38 @@
...
+    # Use Tailscale domain
+    if [ -z "${TAILSCALE_DOMAIN}" ]; then
+        echo "Error: TAILSCALE_ENABLED is true but TAILSCALE_DOMAIN is not set"
+        exit 1
+    fi
+    CRS_HOSTNAME="${CRS_HOSTNAME:-buttercup-ui}"
+    UI_URL="http://${CRS_HOSTNAME}.${TAILSCALE_DOMAIN}"
+else
+    # Use local port-forward
+    kubectl port-forward -n "${BUTTERCUP_NAMESPACE:-crs}" service/buttercup-ui 31323:1323 &
```
hbrodin, line 27: Just occurred to me that we might want to record the pid for this case and have the `web-ui.sh` wait for it or, otherwise kill it on trap (e.g. Ctrl+C) to ensure we don't leave dangling port-forwards. What do you think? Overkill?

hbrodin (approved):
Looks good! I left a suggestion that we could consider.

scripts/web_ui.sh
ret2libc: I'm not sure that's possible, as the script terminates and the port-forwarding should still be up and running to reach the webui.

hbrodin: That's my point, keep the script running waiting for the port-forward to end, or for the user to kill this script and then kill the port-forward.

ret2libc: I see. We can do that yeah! i'll fix it asap

scripts/web_ui.sh
```diff
@@ -23,16 +23,37 @@ if [ "${TAILSCALE_ENABLED:-false}" = "true" ]; then
...
-else
-    echo "Please open $UI_URL in your browser."
+    echo "Opening web UI at $UI_URL..."
+    if command -v xdg-open >/dev/null 2>&1; then
+        xdg-open "$UI_URL"
+    elif command -v open >/dev/null 2>&1; then
+        open "$UI_URL"
+    else
+        echo "Please open $UI_URL in your browser."
+    fi
```
hbrodin: I think you need to restructure this a bit, otherwise it won't show for non-local scenarios.

hbrodin: Other than that it looks good! 👍

ret2libc: oh yes sorry, i still need to test this my computer is busy with another test. will fix this for real in a bit

hbrodin (approved):
Nice!

PR #357, opened 2025-09-12
Add support for C++ projects

ret2libc:
* Split C and C++ projects
* Add support for C++ files in tree-sitter
* Remove patch_validation step because it does not support c++ and it was anyway a leftover from the AIxCC competition. It can actually happen that a patch needs to fix a harness and checking if the patch is only in the right type of file was never that good.
* Remove language-identifier binaries


patcher/src/buttercup/patcher/agents/tools.py
```diff
@@ -183,11 +183,21 @@ def get_function_tool_impl(function_name: str, file_path: str | None, state: Bas
...
     otherwise pass None. Use this when you want to get information about a \
     function. If not sure about the file path, pass None. Prefer using this \
     tool over any other and rely on others only if this tool fails or does \
-    not work.
+    not work. If the function is a method, just pass the method name and, if
+    known, the file_path as usual. Do NOT pass the class name as part of the
+    function name.
+
+    Do NOT do this:
+    - `get_function("MyClass::add", "/src/example_project/test.cpp")`
```
hbrodin, line 195: How can we qualify based on class? Is that not possible? Seems like it would be useful.

ret2libc: We can't, that's correct... I guess we could do that but I'd create a separate issue/PR for that because it affects both Java and C++. It was not implemented even for that AFAIS.

ret2libc:
@hbrodin I've tested libraw library, which is C++. It worked :) That's the only C++ project I've tested so far.

patcher/src/buttercup/patcher/agents/tools.py
hbrodin: Created #361.

PR #359, opened 2025-09-12
Allow to disable find_tests step while developing

ret2libc:


patcher/src/buttercup/patcher/config.py
```diff
@@ -19,6 +19,7 @@ class ProcessCommand(BaseModel):
     sanitizer: CliPositionalArg[str] = Field(description="Sanitizer")
     crash_input_path: CliPositionalArg[str] = Field(description="Crash Input Path")
     stacktrace_path: CliPositionalArg[str] = Field(description="Stacktrace Path")
+    disable_find_tests: CliImplicitFlag[bool] = Field(default=False, description="Disable find tests")
```
hbrodin: I guess another option if you want to get rid of negation is to implement `find_tests` with default value of `True`.

hbrodin (approved):
LGTM! (One potential change, but really a matter of taste)

patcher/src/buttercup/patcher/config.py
ret2libc: Yes indeed, I just realized that pydantic-settings also adds automatically the `--no-<option>` flag, so that should be better!

PR #363, opened 2025-09-29
deployment: fix bitnami deprecation errors

ret2libc:
The bitnami/kubectl image was failing to be downloaded... I guess that might have to do with the fact they are retiring/moving some bitnami images. We do not need anything special here, just kubectl, so we can switch to alpine.


hbrodin (approved):
LGTM!

PR #366, opened 2025-10-01
deployment: remove a few warnings during deployment

ret2libc:
There were a few warnings like:
```
Warning: spec.template.spec.containers[1].env[21]: `NODE_DATA_DIR` hides previous definition
```
because the same env vars were included multiple times.

### DIFF
--- a/deployment/k8s/charts/competition-api/values.yaml
+++ b/deployment/k8s/charts/competition-api/values.yaml
@@ -15,4 +15,3 @@ resources:
 
 nodeSelector: {}
 tolerations: []
-affinity: {}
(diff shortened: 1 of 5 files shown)


deployment/k8s/charts/competition-api/values.yaml
```diff
@@ -15,4 +15,3 @@ resources:
 
 nodeSelector: {}
 tolerations: []
-affinity: {}
```
hbrodin, line 18: Just out of curiosity, why this change?

hbrodin (approved):
LGTM!

deployment/k8s/charts/competition-api/values.yaml
ret2libc: probably unnecessary, i 'm also seeing a few "affinity" warnings while deployment and thought this was one of those. i think it wasn't but, this field was not used anyway in any of the templates so I just deleted it. Does it really affect anything? I thought that things not included in the templates were not really used.

The affinity warnings are probably due to some dependency chart we use, like signoz etc.

hbrodin: I don't expect any issues from removing this. Was mostly curious if you'd seen any difference.

PR #367, opened 2025-10-03
Use our modified oss-fuzz

ret2libc:
libpng has changed the default branch to libpng18, but that doesn't contain the oss-fuzz files. Let's use our own fork of oss-fuzz for now, until https://github.com/google/oss-fuzz/pull/14080 is ready.


michaelbrownuc (approved):
LGTM, we can revert when the PR is resolved.

PR #373, opened 2025-10-17
scripts: auto-install a few extra tools for AKS deployment

ret2libc:


ret2libc:
@hbrodin not really, i just followed the existing code

PR #378, opened 2025-10-31
Fix ARM64 crash reproduction and helper.py patching

hbrodin:
The tracer bot failed to reproduce crashes on ARM64 due to incorrect parameter passing in the reproduce_impl patch. The err_result value was being passed as a positional argument to docker_run(), which interpreted it as print_output=False, redirecting all output to /dev/null.

Changes:
- Fix reproduce_impl to pass architecture as keyword arg only
- Refactor ARM64 patching into separate dockerfile and runner functions
- Add comprehensive helper.py patching for ARM64:
  * Patch image_name variables to use :manifest-arm64v8 tag
  * Patch BASE_RUNNER_IMAGE assignment with tag stripping
  * Fix debug mode tag insertion (insert -debug before tag, not after)
  * Prevent double-tagging in _get_base_runner_image()
- Default CLI architecture parameters to ARCHITECTURE constant

This enables the tracer bot to see fuzzer output and successfully detect crash reproduction on ARM64 systems.

### DIFF
--- a/deployment/k8s/values-upstream-minikube.template
+++ b/deployment/k8s/values-upstream-minikube.template
@@ -51,7 +51,7 @@ redis:
       enabled: false
 
 build-bot:
-  replicaCount: 1
+  replicaCount: 3
 
 litellm:
   masterKey: "${LITELLM_MASTER_KEY}"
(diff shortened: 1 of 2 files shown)


hbrodin:
~~@ret2libc still running some local tests. Will update the PR once done.~~  ✅

ret2libc:
Oops, I forgot to mention this last time we talked indeed: https://github.com/google/oss-fuzz/pull/14178

ret2libc:
Seems to be working for me! How did it go on your end @hbrodin ?

hbrodin:
I've successfully run it so I think it works. Sorry, forgot to update the issue 🫣 .

PR #384, opened 2025-11-05
fuzzer: demangle function names when storing Coverage info

ret2libc:
Without this fix, the coverage-bot was using mangled names and the seed-gen was retrieving those, making the job of targeting a function harder.

### DIFF
--- a/common/tests/test_challenge_task.py
+++ b/common/tests/test_challenge_task.py
@@ -1,5 +1,6 @@
 import base64
 import os
+import platform
 import subprocess
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


fuzzer/src/buttercup/fuzzing_infra/coverage_runner.py
```diff
@@ -49,7 +50,15 @@ def _process_function_coverage(coverage_data: dict[str, Any]) -> list[CoveredFun
                 if "name" not in function or "regions" not in function:
                     continue
 
+                # Demangle the function name if necessary, depending on the language.
                 name = function["name"]
+                if self.language in {Language.CPP, Language.C}:
+                    try:
+                        demangled_name: str = cxxfilt.demangle(name)
+                        name = demangled_name
```
hbrodin: Is there ever a risk that name would be empty once it is demangled? If so, we would add `CoveredFunction` with empty names later on. Not sure what the consequences of that would be. I'm not exactly clear on when/why this would happen, but we could consider a fallback. I.e. using the `name` instead of `demanged_name` if `demangled_name == ""`

hbrodin (approved):
Looks good! Nice catch! Please just consider my comment. Not sure it is an issue, but at least wanted to highlight the possibility.

hbrodin:
Looks good!

PR #396, opened 2025-11-25
Create CODEOWNERS

smoelius:
@ret2libc Could you please review? Does this file look right to you?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,7 @@
 * @ret2libc @hbrodin @michaelbrownuc
-seed-gen/fuzzer @reytchison 
+seed-gen @reytchison
+fuzzer @reytchison
+fuzzer_runner @reytchison 
 program-model @evandowning 
 patcher @ret2libc
 orchestrator @hbrodin


hbrodin:
LGTM, but I'll let @ret2libc weigh in.

CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+* @ret2libc @hbrodin @michaelbrownuc
+seed-gen/fuzzer @reytchison
```
reytchison: I think this change is necessary so we aren't matching a subdirectory `seed-gen/fuzzer`? Also added another fuzzer dir.
```suggestion
seed-gen @reytchison
fuzzer @reytchison
fuzzer_runner @reytchison 
```

smoelius: I'm happy to commit whatever changes are desired.

reytchison: Ok, could we commit this change, please?

reytchison (approved):
LGTM, at least for what I'm the codeowner of.

PR #397, opened 2025-11-26
Fix CRS_KEY_TOKEN_HASH truncation in setup-local

hbrodin:
- auth_tool.py: Use plain print() when piped; Rich wraps at 80 chars
  - common.sh: Escape $ in hash before sed (interpreted as backreferences)

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -624,9 +624,7 @@ generate_crs_key_id_token() {
 
     portable_sed "s|.*export CRS_KEY_ID=.*|export CRS_KEY_ID=\"$crs_key_id\"|" deployment/env
     portable_sed "s|.*export CRS_KEY_TOKEN=.*|export CRS_KEY_TOKEN=\"$crs_key_token\"|" deployment/env
-    # Escape $ characters in the hash so sed treats them as literals (Argon2id hashes contain multiple $)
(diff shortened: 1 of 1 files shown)


scripts/common.sh
```diff
@@ -624,7 +624,9 @@ generate_crs_key_id_token() {
 
     portable_sed "s|.*export CRS_KEY_ID=.*|export CRS_KEY_ID=\"$crs_key_id\"|" deployment/env
     portable_sed "s|.*export CRS_KEY_TOKEN=.*|export CRS_KEY_TOKEN=\"$crs_key_token\"|" deployment/env
-    portable_sed "s|.*export CRS_KEY_TOKEN_HASH=.*|export CRS_KEY_TOKEN_HASH='$crs_key_token_hash'|" deployment/env
+    # Escape $ characters in the hash so sed treats them as literals (Argon2id hashes contain multiple $)
```
ret2libc: is this really necessary? I didn't have problems usually with this because wrapped in `'`.

hbrodin: I did the other change after this so I can double check if that alone fixed it. But, I do recall the value being cut in half without this. Will check.

hbrodin: You were absolutely right, good catch!

PR #401, opened 2025-12-11
Improve coverage tracking precision

### LINKED ISSUE #400: Coverage metric precision suboptimal
Problem

  The coverage-bot counts all LLVM coverage regions as reachable code, but many region types don't represent executable code:

  - SkippedRegions: Code inside unmet #ifdef conditions (never compiled)
  - GapRegions: Whitespace/formatting artifacts
  - BranchRegions: Metadata that overlaps with existing CodeRegions
  - ExpansionRegions: Macro call markers (actual code stored separately)

  The region kind is available at index 7 of each region array in llvm-cov export output but is currently ignored.

  Impact

  1. Inflated total line counts → artificially low coverage percentages
  2. Macro coverage ignored → expanded code not tracked at all
  3. Platform-specific code skews metrics → #ifdef WINDOWS code counted on Linux builds

hbrodin:
Previously coverage tracking included all regions. This changes coverage tracking to be more precise.

Macros that result in code are represented as a single line in the function, and any executed line in the macro will make that line 'covered'. Beyond that, only CodeRegions will count. This will prevent code guarded by #if that aren't in the binary from being considered reachable for example. The idea is that seed-gen would have more accurate information when selecting functions to target.

I've done some testing and it appears as if the new implementation in general reaches more lines and covers more functions.

Closes #400.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,11 +132,11 @@ send-integration-task:
 		echo "Error: CRS namespace not found. Deploy first with 'make deploy'."; \
 		exit 1; \
 	fi
-	kubectl port-forward -n $${BUTTERCUP_NAMESPACE:-crs} service/buttercup-ui 31323:1323 &
-	@sleep 3
-	./orchestrator/scripts/task_integration_test.sh
-	pkill -f "kubectl port-forward" || true
-	exit 0
+	@kubectl port-forward -n $${BUTTERCUP_NAMESPACE:-crs} service/buttercup-ui 31323:1323 & \
+	PORT_FORWARD_PID=$$!; \
+	sleep 3; \
+	./orchestrator/scripts/task_integration_test.sh; \
+	kill $$PORT_FORWARD_PID 2>/dev/null || true
 
 send-libpng-task:
 	@echo "Running libpng task..."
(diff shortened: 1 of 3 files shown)


fuzzer/docs/COVERAGE_REGION_TYPES.md
```diff
@@ -0,0 +1,111 @@
...
+
+Standard executable code. All lines in the region are counted as reachable. Lines with `execution_count > 0` are counted as covered.
+
+### ExpansionRegion (kind=1)
+
+Macro call sites. These are pointers to expanded content, not executable code themselves. The expanded content's CodeRegions are stored separately in `files[].expansions[].target_regions`.
+
+**Our approach**: Count the macro call as ONE reachable line (the start line of the macro invocation). If ANY CodeRegion in the expansion's target_regions has `execution_count > 0`, count that line as covered.
+
+**Rationale**: This avoids inflating line counts with macro internals while still tracking whether macro code was exercised.
```
ret2libc: What about macros that have `if` inside? By counting the whole macro as 1, we don't see the difference between inputs that actually cover more code. Right?

hbrodin: The other option is to inline the row counts. Would you prefer that instead? I tried to keep it simple as looking at the function definition it becomes harder to reason about coverage on specific lines when we only see the "macro call". If you think it'd be more valuable it shouldn't be too hard to achieve.

hbrodin: Additional note: We would previously also have counted the macro as one line, regardless of if it expanded to any line or many lines.

reytchison: Nice work improving the coverage map! It would be nice to count expanded macro lines, including for other cases like if 1 macro expands to many lines but another macro expands to 1 line. However I'm OK with noting this and considering it as future work since the new approach isn't a regression for how we count macro lines.

ret2libc: Yes, same as what @reytchison said!

hbrodin: Ok, I've updated the branch to recursively expand macros now. It became a bit bigger but it should be pretty accurate now I think.

PR #408, opened 2026-01-04
Fix minikube docker-env eval by forcing Bash output

kevin-valerio:
Force minikube `docker-env` to emit Bash syntax (--shell bash) to prevent (for example) Fish `set -gx` output from breaking eval in Bash scripts. 

This fixes `crs-architecture.sh: line 110: set: -g: invalid option` kind of issues when running Buttercup on non-Bash shells.

### DIFF
--- a/deployment/crs-architecture.sh
+++ b/deployment/crs-architecture.sh
@@ -107,7 +107,7 @@ up() {
 				minikube status
 
 				echo -e "${BLU}Building local docker images${NC}"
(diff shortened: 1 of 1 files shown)


kevin-valerio:
> LGTM. I haven't seen this when running on my mac though, but I assume it is something you've experienced?

Indeed (that was a Fish issue more than a Mac issue)

PR #410, opened 2026-01-08
feat: add extract_povs command to buttercup-util

hbrodin:
Add new CLI subcommand to extract PoVs, stack traces, and patches from Redis submissions into a structured directory format for easy analysis.

Features:
- Extracts crash inputs (PoV files) via kubectl cp from cluster pods
- Writes fuzzer and tracer stack traces to text files
- Exports associated patches with metadata
- Organizes output by project/task_id/vulnerability
- Supports filtering by task_id and passed_only options
- Skips empty patch trackers (placeholders that never received content)

### DIFF
--- a/common/src/buttercup/common/util_cli.py
+++ b/common/src/buttercup/common/util_cli.py
@@ -1,4 +1,6 @@
+import json
 import logging
+import subprocess
 from pathlib import Path
 from typing import Annotated
 from uuid import uuid4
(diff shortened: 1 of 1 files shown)


common/src/buttercup/common/util_cli.py
```diff
@@ -136,6 +149,209 @@ class Config:
         extra = "allow"
 
 
+def get_pod_name(namespace: str, label: str) -> str | None:
```
ret2libc, line 152: aren't there multiple pods matching a label?

hbrodin: well... yes, but we only need one of them as the same dir is mounted in everyone.

ret2libc (approved):

PR #411, opened 2026-01-14
Sanitize exception messages for git clone command

reytchison:


orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
```diff
@@ -84,13 +84,17 @@ def create_challenge_tarball(
...
+                result = subprocess.run(
+                    ["git", "clone", clone_url, sub_path],
+                    capture_output=True,
+                    text=True,
+                    check=True,
+                )
+                logger.info(f"Cloned {repo_url} to {sub_path}")
+            except Exception:
+                sanitized_command = f"git clone {repo_url} {sub_path}"
+                raise Exception(f"Failed to clone repository. Sanitized command: {sanitized_command}")
```
hbrodin: Ideally we would have some feedback for the user what actually happened, however I think that is hard so this is probably the next best result. Maybe we could output the return code? Also, maybe if there is no pat/username I think we could just keep the old one.
Really, up to you.

hbrodin (approved):
LGTM! Added a few comments that you could consider.

orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
reytchison: Updated to not sanitize if there is no PAT, preserving all feedback. I decided to not parse out information from exception (e.g. ret code) for sanitized case for simplicity.

PR #413, opened 2026-01-15
feat: add git lfs support for challenge repositories

hbrodin:
Use git-lfs when downloading challenges to handle large binary files.

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
+++ b/orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
@@ -58,7 +58,7 @@ class ChallengeService:
 
         # git-lfs not installed or not available
         if ls_result.returncode != 0:
-            logger.debug(f"[{context}] git-lfs not available: {ls_result.stderr.strip()}")
+            logger.warning(f"[{context}] git-lfs not available: {ls_result.stderr.strip()}")
             return
 
         # No LFS files in this repo


orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
```diff
@@ -34,6 +34,51 @@ def __init__(self, storage_dir: Path, base_url: str) -> None:
...
+        ls_result = subprocess.run(
+            ["git", "lfs", "ls-files"],
+            cwd=repo_path,
+            capture_output=True,
+            text=True,
+        )
+
+        # git-lfs not installed or not available
+        if ls_result.returncode != 0:
+            logger.debug(f"[{context}] git-lfs not available: {ls_result.stderr.strip()}")
```
reytchison: Nit: missing a dependency which could break cloning a challenge should be a higher log level 
```suggestion
            logger.warning(f"[{context}] git-lfs not available: {ls_result.stderr.strip()}")
```

orchestrator/test/test_challenge_service.py
```diff
@@ -431,3 +431,117 @@ def test_create_sarif_broadcast_complex_sarif(self, challenge_service):
...
+    def test_no_lfs_files_in_repo(self, mock_run, challenge_service, mock_repo_path):
+        """Test that git lfs pull is not called when no LFS files exist."""
+        # git lfs ls-files returns empty (no LFS files)
+        mock_run.return_value = MagicMock(returncode=0, stdout="", stderr="")
+
+        challenge_service._pull_lfs_files(mock_repo_path, "test-repo@main")
+
+        # Should only call git lfs ls-files, not git lfs pull
+        assert mock_run.call_count == 1
+        mock_run.assert_called_once_with(
```
reytchison, line 481: This test case should also check that git lfs pull isn't called?

reytchison: (Based on the comments)

hbrodin: It is a bit hidden, but it actually does. `call_count ==1` only allows for one subprocess call, and the `assert_called_once_with` ensures it is the `git lfs ls-files` one.

PR #415, opened 2026-01-20
Use project specified in oss-fuzz

hbrodin:
This will use the contents of the oss-fuzz provided container as the projects contents instead of a user supplied git url. This will allow to run projects very similar to how oss-fuzz does it. It will enable analysis of e.g. libmodbus who was previously not possible due to the fuzzing harnesses being overwritten/shadowed by the mount of project directory.

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
+++ b/orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
@@ -298,7 +298,11 @@ class ChallengeService:
         except FileNotFoundError:
             logger.warning(f"Dockerfile not found at {dockerfile_path}, using default WORKDIR")
             workdir = default_workdir
-        focus_dir = workdir.parts[-1]  # e.g., /src/libxml2 -> libxml2
+        # If WORKDIR is just /src, use project_name as focus_dir
+        if workdir == Path("/src"):
+            focus_dir = project_name
+        else:
+            focus_dir = workdir.parts[-1]  # e.g., /src/libxml2 -> libxml2
         logger.info(f"Parsed WORKDIR: {workdir}, focus_dir: {focus_dir}")
 
         # 2. Build the Docker image using OSS-Fuzz's helper.py


hbrodin:
Still analyzing to make sure nothing breaks.

hbrodin:
I think it is ready for review.

orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
```diff
@@ -258,10 +245,157 @@ def _extract_focus_dir(self, repo_url: str) -> str:
...
+                if match:
+                    workdir = match.group(1).replace("$SRC", "/src")
+                    workdir = Path(workdir)
+                    if not workdir.is_absolute():
+                        workdir = Path("/src") / workdir
+                    return workdir
+        except FileNotFoundError:
+            logger.warning(f"Dockerfile not found at {dockerfile_path}, using default WORKDIR")
+
+        return default
```
ret2libc: i think we have this function elsewhere in the codebase

orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
```diff
@@ -258,10 +245,157 @@ def _extract_focus_dir(self, repo_url: str) -> str:
...
+        # Calculate SHA256 hash
+        sha256_hash = self._calculate_sha256(tarball_path)
+        # Rename the tarball to use the sha256 hash
+        new_tarball_path = self.storage_dir / f"{sha256_hash}.tar.gz"
+        os.rename(tarball_path, new_tarball_path)
+
+        logger.info(f"Created tarball: {new_tarball_path}")
+        return sha256_hash
+
     def create_task_for_challenge(
```
ret2libc, line 442: not completely sure but i think this function can be simplified a bit

orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
hbrodin: Yes, we do. Using it now.

orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
hbrodin: Did a refactor, let me know what you think.

orchestrator/src/buttercup/orchestrator/ui/competition_api/services/challenge_service.py
```diff
@@ -258,10 +227,218 @@ def _extract_focus_dir(self, repo_url: str) -> str:
...
+        # 1. Parse WORKDIR from Dockerfile
+        dockerfile_path = oss_fuzz_path / "projects" / project_name / "Dockerfile"
+        default_workdir = Path("/src") / project_name
+        try:
+            lines = dockerfile_path.read_text().splitlines()
+            workdir = ChallengeTask._workdir_from_lines(lines, default=default_workdir)
+        except FileNotFoundError:
+            logger.warning(f"Dockerfile not found at {dockerfile_path}, using default WORKDIR")
+            workdir = default_workdir
+        focus_dir = workdir.parts[-1]  # e.g., /src/libxml2 -> libxml2
```
ret2libc: is this always the case?

PR #416, opened 2026-01-20
Improve fuzzer logging

hbrodin:
To help when integrating new harnesses with buttercup, this provides more detailed logging about harness execution and exit.

### DIFF
--- a/fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
+++ b/fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
@@ -91,13 +91,38 @@ class RunnerProxy:
 
             try:
                 stdout, stderr = process.communicate(timeout=subprocess_timeout)
-                if process.returncode != 0:
+
+                # Log the return code with interpretation
+                returncode = process.returncode
+                if returncode != 0:
+                    # Interpret common exit codes
+                    if returncode > 128:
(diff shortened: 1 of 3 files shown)


fuzzer/src/buttercup/fuzzing_infra/runner_proxy.py
```diff
@@ -91,13 +91,48 @@ def _run_subprocess_task(self, cmd: list[str], timeout: int, task_type: str) ->
...
+                            11: "SIGSEGV",
+                            13: "SIGPIPE",
+                            14: "SIGALRM",
+                            15: "SIGTERM",
+                        }
+                        signal_name = signal_names.get(signal_num, f"signal {signal_num}")
+                        logger.warning(
+                            f"{task_type} process terminated by {signal_name} "
+                            f"(exit code {returncode} = 128 + {signal_num})"
+                        )
```
ret2libc, line 108: can we reuse something from https://docs.python.org/3/library/signal.html ?

hbrodin: That's a great idea! I'll check!

hbrodin: Updated the implementation.

PR #417, opened 2026-01-21
Fixes to docker-compose file

ret2libc:


dev/docker-compose/compose.yaml
```diff
@@ -1,13 +1,13 @@
-include:
- - ../../competition-server/compose.yaml
- - ../../competition-server/signoz/compose.yaml
+#include:
+# - ../../competition-server/compose.yaml
+# - ../../competition-server/signoz/compose.yaml
```
hbrodin: Why comment?

ret2libc: We could remove 🤷 maybe i should move these services under a "profile"

hbrodin: Yeah, I think that'd be better.

dev/docker-compose/compose.yaml
```diff
@@ -240,15 +184,30 @@ services:
     build:
       context: ../../
       dockerfile: ./seed-gen/Dockerfile
-    command: ["seed-gen", "server"]
+    command: >
```
hbrodin, line 187: I don't follow what this does. Probably correct, just that I don't see how it works.

ret2libc: in k8s we use https://github.com/trailofbits/buttercup/blob/main/deployment/k8s/templates/litellm-user-keys-job.yaml to create the litellm keys and then inject them back in k8s secrets. Here we can't do that exactly, so we wait for the `litellm-user-keys-setup` service to create the LITELLM_API_KEY_FILE and then we read the key from there.

PR #461, opened 2026-01-30
Add optional tmpfs storage for fuzzing corpus

hbrodin:
Add support for storing the node-local fuzzing corpus in tmpfs (RAM-based filesystem) for improved I/O performance during fuzzing operations.

Key changes:
- Add CORPUS_TMPFS_PATH environment variable to configure tmpfs location
- Add cross-filesystem safe operations (copy+delete fallback for EXDEV)
- Update Corpus class to use tmpfs path when enabled while maintaining correct remote path calculation for rsync synchronization
- Add Helm chart configuration using /dev/shm (requires no host setup)
- Update fuzzer-bot, coverage-bot, merger-bot, seed-gen deployments

The feature is disabled by default. Enable by setting global.volumes.corpusTmpfs.enabled=true in values.yaml.

Uses hostPath to /dev/shm/buttercup-corpus which is available on Linux systems without requiring any Kubernetes host configuration.

### DIFF
--- a/common/src/buttercup/common/corpus.py
+++ b/common/src/buttercup/common/corpus.py
@@ -166,6 +166,7 @@ class InputDir:
                 continue
             file_path = os.path.join(path, file)
             try:
+                assert os.path.isfile(file_path)
                 with open(file_path, "rb") as f:
                     hash_filename = hash_file(f)
                 dst_path = os.path.join(path, hash_filename)
--- a/common/src/buttercup/common/node_local.py
+++ b/common/src/buttercup/common/node_local.py
@@ -89,7 +89,12 @@ def rename_atomically(src: Path, dst: Path) -> Path | None:
 
 
 def _copy_and_delete(src: Path, dst: Path) -> Path | None:
-    """Copy a file/directory and delete the source. Used for cross-filesystem operations."""
+    """Copy a file/directory and delete the source. Used for cross-filesystem operations.
+
+    Note: Unlike os.rename, this intentionally returns None if dst already exists
+    rather than overwriting. This supports the "first pod wins" pattern used by
+    callers in distributed scenarios where concurrent pods may race to create the same file.
+    """
     try:
         if src.is_dir():
             # For directories, use copytree
@@ -114,35 +119,6 @@ def _copy_and_delete(src: Path, dst: Path) -> Path | None:
     except Exception as e:
         logger.error(f"Failed to copy {src} to {dst}: {e}")
         raise
-
-
-def move_file_cross_fs(src: Path | str, dst: Path | str) -> Path:
-    """Move a file, handling cross-filesystem operations.
-
-    Unlike os.rename, this works across filesystem boundaries by falling back
-    to copy+delete when necessary.
-
-    Args:
-        src: Source file path
-        dst: Destination file path
-
-    Returns:
-        The destination path
-    """
-    src = Path(src)
-    dst = Path(dst)
-    dst.parent.mkdir(parents=True, exist_ok=True)
-
-    try:
-        os.rename(src, dst)
-    except OSError as e:
-        if e.errno == errno.EXDEV:
-            # Cross-filesystem move - copy then delete
-            shutil.copy2(src, dst)
-            os.unlink(src)
-        else:
-            raise
-    return dst
 
 
 def remote_path(local_path: NodeLocalPath) -> RemotePath:
--- a/common/tests/test_corpus.py
+++ b/common/tests/test_corpus.py
@@ -461,22 +461,6 @@ class TestGetCorpusStoragePath:
 class TestCrossFilesystemOperations:
     """Tests for cross-filesystem file operations."""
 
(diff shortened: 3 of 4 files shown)


docs/tmpfs-corpus.md
reytchison, line 1: Is this intended to be added? Some information is useful for describing the in-memory corpus, but some of the information is too specific to implementing the PR. Also no other docs exist in `docs/`.

deployment/k8s/values.yaml
```diff
@@ -81,6 +81,23 @@ global:
...
+    # Optional tmpfs volume for fuzzing corpus storage (per-node, shared across pods)
+    # When enabled, corpus data is stored in RAM for improved I/O performance
+    #
+    # Default uses /dev/shm which is:
+    #   - Available on all Linux systems (no host configuration needed)
+    #   - Already mounted as tmpfs
+    #   - Sized to ~50% of RAM by default
+    #   - Shared with other processes using POSIX shared memory
+    #
+    # For dedicated/isolated tmpfs, mount one on each node and change hostPath.
```
reytchison: Nit: Comment is a bit verbose

docs/tmpfs-corpus.md
hbrodin: 😊 No...

deployment/k8s/values.yaml
hbrodin: Agree. Made it shorter.

common/src/buttercup/common/node_local.py
```diff
@@ -57,13 +77,74 @@ def rename_atomically(src: Path, dst: Path) -> Path | None:
...
     return dst
 
 
+def _copy_and_delete(src: Path, dst: Path) -> Path | None:
+    """Copy a file/directory and delete the source. Used for cross-filesystem operations."""
+    try:
+        if src.is_dir():
+            # For directories, use copytree
+            if dst.exists():
+                logger.debug(f"Destination {dst} already exists, skipping copy")
```
reytchison, line 102: When handling if dst already exists, I think we should match the semantics of `os.rename` since we're using this as a fallback for `os.rename`. From the [docs](https://docs.python.org/3/library/os.html#os.rename):
> On Unix, if src is a file and dst is a directory or vice-versa, an [IsADirectoryError](https://docs.python.org/3/library/exceptions.html#IsADirectoryError) or a [NotADirectoryError](https://docs.python.org/3/library/exceptions.html#NotADirectoryError) will be raised respectively. If both are directories and dst is empty, dst will be silently replaced. If dst is a non-empty directory, an [OSError](https://docs.python.org/3/library/exceptions.html#OSError) is raised. If both are files, dst will be replaced silently if the user has permission.

common/src/buttercup/common/node_local.py
```diff
@@ -57,13 +77,74 @@ def rename_atomically(src: Path, dst: Path) -> Path | None:
...
+                return None
+            shutil.copy2(src, dst)
+            os.unlink(src)
+        return dst
+    except Exception as e:
+        logger.error(f"Failed to copy {src} to {dst}: {e}")
+        raise
+
+
+def move_file_cross_fs(src: Path | str, dst: Path | str) -> Path:
```
reytchison: I think this is unused?

common/src/buttercup/common/node_local.py
reytchison: We could use `shutil.move` like `InputDir` does. However we'd want to check whether source and dest are directories to preserve the `os.rename` semantics.

common/src/buttercup/common/corpus.py
```diff
@@ -112,7 +168,9 @@ def hash_corpus(cls, path: str) -> list[str]:
             try:
                 with open(file_path, "rb") as f:
                     hash_filename = hash_file(f)
-                os.rename(file_path, os.path.join(path, hash_filename))
+                dst_path = os.path.join(path, hash_filename)
+                # Use shutil.move which handles cross-filesystem operations
+                shutil.move(file_path, dst_path)
```
reytchison, line 174: Confirm the source is a file so that we're following `os.rename` semantics
```suggestion
                assert os.path.isfile(file_path)
                shutil.move(file_path, dst_path)
```

common/tests/test_corpus.py
```diff
@@ -302,3 +303,271 @@ def test_copy_file_with_remote(temp_dir):
...
+    def test_move_file_cross_fs_same_filesystem(self, temp_dir):
+        """Test move_file_cross_fs works on same filesystem."""
+        src = Path(temp_dir) / "src_file"
+        dst = Path(temp_dir) / "subdir" / "dst_file"
+
+        # Create source file
+        src.write_bytes(b"test content")
+
+        # Move file
+        result = node_local.move_file_cross_fs(src, dst)
```
reytchison: There are tests for `node_local.move_file_cross_fs` in two files: this one and test_cross_filesystem.py

reytchison (approved):

PR #542, opened 2026-05-11
Security: Path Traversal Vulnerability in File Download

tomaioo:
## Summary

Security: Path Traversal Vulnerability in File Download

## Problem

**Severity**: `High` | **File**: `orchestrator/src/buttercup/orchestrator/utils.py:L7`

The `response_stream_to_file` function in `orchestrator/src/buttercup/orchestrator/utils.py` accepts an optional `filepath` parameter that could allow path traversal attacks if user-controlled input is passed. An attacker could potentially write files outside the intended directory by manipulating the filepath.

## Solution

Add path validation to ensure the filepath stays within the intended directory. Use os.path.realpath() to resolve symlinks and validate the final path is within an allowed directory.

## Changes

- `orchestrator/src/buttercup/orchestrator/utils.py` (modified)

### DIFF
--- a/orchestrator/src/buttercup/orchestrator/utils.py
+++ b/orchestrator/src/buttercup/orchestrator/utils.py
@@ -1,10 +1,20 @@
 import hashlib
+import os
 from pathlib import Path
 
 import requests
 
 
-def response_stream_to_file(session: requests.Session, url: str, filepath: Path | None = None) -> str:
+def response_stream_to_file(session: requests.Session, url: str, filepath: Path | None = None, base_dir: Path | None = None) -> str:
+    if filepath:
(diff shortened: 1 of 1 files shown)


tomaioo:
Thanks for addressing this path traversal vulnerability. The proposed fix using `os.path.realpath()` to validate the resolved path stays within the allowed directory is the right approach. The CLA status appears to be updating automatically—your signature should propagate shortly.

tomaioo:
Thanks for addressing this path traversal vulnerability. The proposed fix using `os.path.realpath()` to validate the resolved path stays within the allowed directory is the right approach. The CLA status appears to be updating automatically—your signature should propagate shortly.

PR #544, opened 2026-05-14
feat(llm): add newer Anthropic + OpenAI models (opus 4.6/4.7, sonnet 4.6, haiku 4.5, gpt-5.4/5.4-mini/5.5)

ret2libc:
## Summary

Add 7 newer model entries to the LiteLLM proxy config (`litellm/litellm_config.yaml`), the k8s Helm values (`deployment/k8s/values.yaml`), and the `ButtercupLLM` enum (`common/src/buttercup/common/llm.py`):

- **Anthropic**: `claude-4.5-haiku`, `claude-4.6-sonnet`, `claude-4.6-opus`, `claude-4.7-opus`
- **OpenAI**: `openai-gpt-5.4-mini`, `openai-gpt-5.4`, `openai-gpt-5.5`

No existing call sites or defaults are changed — the new enum members are available for opt-in adoption without forcing migration of `OPENAI_GPT_4_1` / `CLAUDE_4_5_SONNET` callers.

## Test plan

- [x] All three sources (enum, litellm config, k8s values) agree on the same 25 model names (no drift).
- [x] `docker compose up -d litellm-db litellm` boots LiteLLM v1.57.8 cleanly with the new config — `/health/readiness` returns `{"status":"connected","db":"connected"}`.
- [x] `GET /v1/models` lists all 25 models including the 7 new entries.
- [x] `POST /v1/chat/completions` with `claude-4.6-opus` is dispatched to api.anthropic.com (provider returns a real Anthropic `request_id`, rejecting only the placeholder key).
- [x] `POST /v1/chat/completions` with `openai-gpt-5.5` is dispatched to api.openai.com (provider rejects only the placeholder key).
- [x] No `ERROR`, `unknown model`, or "not supported" entries in litellm boot logs.
- [x] `docker compose down -v` cleans up with no residual containers or volumes.


litellm/litellm_config.yaml
```diff
@@ -108,6 +129,30 @@ model_list:
       api_key: os.environ/ANTHROPIC_API_KEY
       tpm: 400000
 
+  - model_name: claude-4.5-haiku
+    litellm_params:
+      model: anthropic/claude-haiku-4-5-20251001
```
hbrodin, line 134: Worth flagging the naming asymmetry for future maintainers: Haiku 4.5 keeps the dated suffix (`claude-haiku-4-5-20251001`) while the 4.6/4.7 Sonnet/Opus entries below use the dateless form (`claude-sonnet-4-6`, `claude-opus-4-6`, `claude-opus-4-7`). That matches Anthropic's documented rule that the dateless format becomes canonical starting with the 4.6 generation; pre-4.6 models still require the dated snapshot. So when the next pre-4.6-era model gets added, it'll need the dated suffix, and when 4.8+ arrives it stays dateless.

hbrodin (approved):
All seven new model IDs verified against current Anthropic deprecation and OpenAI model docs; the 3-way consistency (enum / litellm config / k8s values) holds and TPM/RPM follow the existing per-variant convention.

PR #545, opened 2026-05-14
docs(skills): add Langfuse agent skills

ret2libc:
## Summary

- Adds `.claude/skills/langfuse/` — vendored verbatim from [github.com/langfuse/skills](https://github.com/langfuse/skills) (MIT). Covers platform-level usage: `langfuse-cli` via npx, docs retrieval via `llms.txt`, and reference guides for instrumentation, prompt migration, SDK upgrades, error analysis, and user-feedback scoring.
- Adds `.claude/skills/buttercup-langfuse/` — project-specific wiring skill: documents `buttercup.common.llm.get_langfuse_callbacks()`, the canonical `RunnableConfig(callbacks=..., tags=..., metadata=...)` pattern already used in `patcher/.../leader.py` and the seed-gen tasks, and the Helm/values surface (`global.langfuse.*`, `langfuse-secret.yaml`, `common-env.yaml`). Defers platform questions to the upstream `langfuse` skill so the two trigger on disjoint intents.

The two skills are complementary: upstream covers "how to use Langfuse generally," the project skill covers "how this repo wires Langfuse."

## Test plan

- [ ] `/reload-plugins` in Claude Code (or restart session) so both skills are picked up
- [ ] Trigger upstream skill by asking a generic Langfuse question (e.g. "how do I query traces filtered by score")
- [ ] Trigger Buttercup skill by asking a repo-specific question (e.g. "I'm adding a new LangGraph entry point in patcher — how do I wire Langfuse?")
- [ ] Confirm no overlap firing: questions clearly in one scope should activate only that skill

## Notes

- The vendored `langfuse/` directory is a clean copy of upstream at the time of this PR (no edits). If we want to track upstream updates we should add a pin file or switch to `npx skills add`; happy to do either as a follow-up.
- One sub-threshold reviewer finding (score 75/100, under the 80 bar): `buttercup-langfuse/SKILL.md` says `langfuse ~=4.0.1` is "provided transitively to anything that depends on `common`", but it's actually under `[project.optional-dependencies] full`. Consumers need `common[full]` (as patcher/seed-gen do). Easy follow-up if we want to be precise.

### DIFF
--- a/.claude/skills/buttercup-langfuse/SKILL.md
+++ b/.claude/skills/buttercup-langfuse/SKILL.md
@@ -16,7 +16,7 @@ Tracing is always optional at runtime: if the env vars are unset or the host is 
 - Defined in `common/src/buttercup/common/llm.py`
 - `@functools.cache`d — cheap to call repeatedly, but the auth/connectivity probe runs only once per process
 - Returns `[]` when Langfuse is disabled, misconfigured, or unreachable
-- Internally: checks `LANGFUSE_HOST`, then `langfuse_auth_check()` against `LANGFUSE_PUBLIC_KEY` / `LANGFUSE_SECRET_KEY`, then constructs `langfuse.langchain.CallbackHandler()`
+- Internally: runs `is_langfuse_available()` (checks `LANGFUSE_HOST`, then HTTP-probes `/api/public/ingestion`), constructs `langfuse.langchain.CallbackHandler()`, then verifies credentials via `langfuse_auth_check()`
 
 Do **not** instantiate `CallbackHandler` directly or read the env vars yourself. Always go through `get_langfuse_callbacks()`.
 
@@ -88,7 +88,7 @@ Where they're set:
   - Injected into pods via the `buttercup.env.langfuse` helper in `deployment/k8s/templates/common-env.yaml`
   - A new component's Deployment template must include that helper to get traces
 
-Python dependency: `langfuse ~=4.0.1`, declared in `common/pyproject.toml`. Provided transitively to anything that depends on `common` — do not pin it independently in another component.
+Python dependency: `langfuse ~=4.0.1`, declared under `[project.optional-dependencies] full` in `common/pyproject.toml`. Provided via the `[full]` extra of `common`. Components that emit traces depend on `common[full]` (as patcher and seed-gen already do). Don't pin `langfuse` independently — re-use the extra.
 
 ## Debugging "I don't see traces"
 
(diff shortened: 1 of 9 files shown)


.claude/skills/buttercup-langfuse/SKILL.md
```diff
@@ -0,0 +1,109 @@
...
+
+Where they're set:
+- Local docker-compose: `dev/docker-compose/env.template`
+- Local make-based deploy: `deployment/env.template` (gated by `LANGFUSE_ENABLED`)
+- Kubernetes: `deployment/k8s/values.yaml` under `global.langfuse.{enabled,host,publicKey,secretKey}`
+  - Rendered into the `<release>-langfuse-secrets` Secret by `deployment/k8s/templates/langfuse-secret.yaml`
+  - Injected into pods via the `buttercup.env.langfuse` helper in `deployment/k8s/templates/common-env.yaml`
+  - A new component's Deployment template must include that helper to get traces
+
+Python dependency: `langfuse ~=4.0.1`, declared in `common/pyproject.toml`. Provided transitively to anything that depends on `common` — do not pin it independently in another component.
```
hbrodin: `langfuse ~=4.0.1` is not a transitive dependency of `common` — it's declared under `[project.optional-dependencies] full` in `common/pyproject.toml`, so a component that depends on plain `common` won't pull it in. The consuming components (patcher, seed-gen, fuzzer, orchestrator, program-model) all declare `common[full]` precisely to get the extras.

Suggested rewording: _"Provided via the `[full]` extra of `common`. Components that emit traces depend on `common[full]` (as patcher and seed-gen already do). Don't pin `langfuse` independently — re-use the extra."_

The same premise also drives the "Don't add a new `langfuse` direct dependency in a sibling component's `pyproject.toml` — it's already provided via `common`" bullet at L106; both should be reworded together.

.claude/skills/buttercup-langfuse/SKILL.md
```diff
@@ -0,0 +1,109 @@
...
+Tracing is always optional at runtime: if the env vars are unset or the host is unreachable, callbacks resolve to an empty list and code paths are unaffected. Follow the existing pattern — do not introduce a parallel integration.
+
+## The one helper to use
+
+`buttercup.common.llm.get_langfuse_callbacks() -> list[BaseCallbackHandler]`
+
+- Defined in `common/src/buttercup/common/llm.py`
+- `@functools.cache`d — cheap to call repeatedly, but the auth/connectivity probe runs only once per process
+- Returns `[]` when Langfuse is disabled, misconfigured, or unreachable
+- Internally: checks `LANGFUSE_HOST`, then `langfuse_auth_check()` against `LANGFUSE_PUBLIC_KEY` / `LANGFUSE_SECRET_KEY`, then constructs `langfuse.langchain.CallbackHandler()`
```
hbrodin: The described order is inverted: `langfuse_auth_check()` is listed before `CallbackHandler()` construction, but the code constructs the handler first and only then validates credentials.

`common/src/buttercup/common/llm.py:81-86`:
```python
if is_langfuse_available():
    try:
        langfuse_handler = CallbackHandler()
        if langfuse_auth_check():
            ...
            return [langfuse_handler]
```

The Debugging section in this same file (L97-99) already walks the steps in the correct order, so L19 contradicts the rest of the doc.

Suggested rewording: _"Internally: runs `is_langfuse_available()` (checks `LANGFUSE_HOST`, then HTTP-probes `/api/public/ingestion`), constructs `langfuse.langchain.CallbackHandler()`, then verifies credentials via `langfuse_auth_check()`."_

.claude/skills/buttercup-langfuse/SKILL.md
ret2libc: Fixed in 930acd9 — adopted your suggested wording at L91, and reworded the L106 "What NOT to do" bullet to point at `common[full]` instead of "already provided via `common`". Verified `common/pyproject.toml` has `langfuse ~=4.0.1` under `[project.optional-dependencies] full` and that patcher/seed-gen declare `common[full]`.

.claude/skills/buttercup-langfuse/SKILL.md
ret2libc: Fixed in 930acd9 — adopted your suggested wording. Confirmed against `common/src/buttercup/common/llm.py:81-86`: `CallbackHandler()` is constructed first, then `langfuse_auth_check()`.

hbrodin (approved):
LGTM

PR #547, opened 2026-05-14
fix(scripts): make install_uv resilient to missing env helper

ret2libc:
## Summary
- `install_uv` in `scripts/common.sh` unconditionally `source`d `$HOME/.local/bin/env` after the uv installer ran, which breaks setup when that helper file is not created (older uv installers, or when the install path differs). Failure mode: `No such file or directory` on the `source` line.
- Source the helper only when it exists, prepend `$HOME/.local/bin` to `PATH` as a fallback, and fail loudly if `uv` is still unreachable after the install.

## Test plan
- [ ] Run a clean setup on a host without prior uv install — `install_uv` completes and subsequent `uv` invocations succeed.
- [ ] Re-run setup on a host where `uv` is already present — function takes the `already installed` branch unchanged.
- [ ] Simulate an install where `$HOME/.local/bin/env` is absent (e.g. delete it before sourcing) — `install_uv` still puts `uv` on `PATH` instead of erroring.

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -82,8 +82,22 @@ install_uv() {
     print_status "Installing uv..."
     if ! command_exists uv; then
         curl -fsSL https://astral.sh/uv/install.sh | sh
-        print_status "Sourcing uv environment..."
-        source "$HOME/.local/bin/env"
+        print_status "Loading uv into PATH..."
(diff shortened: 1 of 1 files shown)


hbrodin (approved):
LGTM!

PR #552, opened 2026-05-15
feat(scripts): Docker-only end-to-end /e2e command for example-libpng

ret2libc:
## What this adds

A Docker-only end-to-end smoke test of the full Buttercup pipeline against
[example-libpng](https://github.com/tob-challenges/example-libpng) — no
Kubernetes/minikube. Mirrors the milestones in
`.github/workflows/system-integration.yml` but tails `docker compose logs`.

- **`scripts/e2e.sh`** — brings the `dev/docker-compose/` stack up, submits the
  canned libpng `trigger_task`, and waits on the pipeline milestones
  (fuzzer build → POV submitted → POV accepted → seed-gen → patch
  generated/approved/passed → bundle submitted; optional SARIF).
- **`make e2e`** (and `make e2e E2E_ARGS=...`).
- **`.claude/commands/e2e.md`** — `/e2e` slash command wrapper.

Flags: `--budget` (LiteLLM per-user max budget, default $3),
`--task-duration`, `--image-tag` / `BUTTERCUP_IMAGE_TAG`, `--no-pull`,
`--keep-up`, `--skip-wait`, `--sarif`, per-phase timeout overrides.

### Image source

By default the stack runs the prebuilt GHCR images via the
`compose.prebuilt.yaml` overlay (nothing built locally). `--no-pull` skips the
`docker compose pull` and uses already-present images (e.g. locally built and
tagged `ghcr.io/trailofbits/buttercup/*:<tag>`).

### .env handling

`e2e.sh` regenerates `dev/docker-compose/.env` each run. It resolves each
value as **environment → existing `.env` → placeholder**, so manually-set
values (e.g. `LANGFUSE_*`) are preserved across runs instead of being
clobbered with empty/placeholder.

## Dependency / merge ordering

The prebuilt path invokes
`docker compose -f compose.yaml -f compose.prebuilt.yaml`. The
`compose.prebuilt.yaml` overlay is **not** in this PR — it lives on the
separate compose-prebuilt branch/PR. This PR should land **after or together
with** that one; on its own the overlay file must already be present in
`dev/docker-compose/`.

## Scope

e2e tooling only — `.claude/commands/e2e.md`, `Makefile`, `scripts/e2e.sh`.
Independent of the three pipeline fixes surfaced while building this
(buttercup-ui internal port, litellm budget enforcement, patcher task
storage), which are their own separate PRs.

## Validation

This tooling was used to drive the pipeline end-to-end during development:
fuzzer build → POV submitted → POV accepted, through seed-gen and patch
generation, with budget tracking and Langfuse tracing.


ret2libc:
Addressed CI lint/static failures in 5210208 (shellcheck SC2015 in scripts/e2e.sh).

scripts/e2e.sh
```diff
@@ -0,0 +1,496 @@
...
+fi
+
+provider_keys_set=0
+for v in ANTHROPIC_API_KEY OPENAI_API_KEY GEMINI_API_KEY; do
+    val="${!v:-}"
+    if [[ -n "$val" && "$val" != "<INSERT_KEY>" ]]; then
+        provider_keys_set=1
+    fi
+done
+if [[ "$provider_keys_set" -eq 0 ]]; then
```
hbrodin: The pre-flight provider-key check rejects the very `.env` path the error tip on line 154 advertises. The loop on lines 146–151 only reads environment variables (`${!v:-}`); the `prev_env` fallback that consults `.env` doesn't run until lines 168–176 — unreachable once this block has called `exit 1`.

A user who saved keys to `.env` on a prior run and re-invokes without re-exporting hits "No LLM provider key found in env" even though the tip says "or add to ${ENV_FILE} first."

Either move this check below the `prev_env` block, or drop the misleading second half of the tip.

scripts/e2e.sh
```diff
@@ -0,0 +1,496 @@
...
+
+###############################################################################
+# Argument parsing
+###############################################################################
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --budget)            BUDGET="$2"; shift 2 ;;
+        --task-duration)     TASK_DURATION="$2"; shift 2 ;;
+        --task-json)         TASK_JSON="$(cat "$2")"; shift 2 ;;
```
hbrodin: `--task-json FILE` silently falls back to the default libpng payload if the file is unreadable. With `set -u` / `set -o pipefail` but no `set -e` (lines 12–13), a failing `cat` here only prints to stderr and leaves `TASK_JSON` empty. The `[[ -z "$TASK_JSON" ]]` guard at line 283 then runs the canned libpng heredoc — so a user who explicitly passed `--task-json` ends up submitting the default task without any error.

Suggest:

```bash
--task-json)
    [[ -r "$2" ]] || { err "--task-json: cannot read $2"; exit 2; }
    TASK_JSON="$(cat "$2")"; shift 2 ;;
```

scripts/e2e.sh
```diff
@@ -0,0 +1,496 @@
...
+    "fuzz_tooling_ref": "fix-libpng",
+    "fuzz_tooling_project_name": "libpng",
+    "duration": ${TASK_DURATION}
+}
+EOF
+)
+fi
+
+log "Submitting task to buttercup-ui /webhook/trigger_task"
+http_code=$(curl -s -o /tmp/e2e_task_resp.$$ -w '%{http_code}' \
```
hbrodin: `/tmp/e2e_task_resp.$$` is a predictable PID-named file in shared `/tmp` and isn't covered by the EXIT trap — if SIGINT/SIGTERM arrives between this `curl` and the `rm -f` on line 304, the file is leaked. `on_exit` / `teardown` (lines 223–240) only run `dc down -v`. Using `mktemp` (and either extending the trap or removing inside `on_exit`) avoids both the predictable name and the cleanup gap.

scripts/e2e.sh
```diff
@@ -0,0 +1,496 @@
...
+###############################################################################
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --budget)            BUDGET="$2"; shift 2 ;;
+        --task-duration)     TASK_DURATION="$2"; shift 2 ;;
+        --task-json)         TASK_JSON="$(cat "$2")"; shift 2 ;;
+        --image-tag)         IMAGE_TAG="$2"; shift 2 ;;
+        --no-pull)           DO_PULL=0; shift ;;
+        --no-build)          DO_PULL=0; shift ;;   # deprecated alias
```
hbrodin: This script is brand-new in this PR, so there's no prior public surface for `--no-build` to alias from — the "Deprecated" framing (and the matching usage line at line 78) is phantom. Suggest dropping both the case branch and the usage line; `--no-pull` covers the same ground.

scripts/e2e.sh
ret2libc: Fixed in d8ee169. The provider-key check now runs *after* the `prev_env` env→`.env` fallback, so a key saved to `.env` on a prior run is accepted, and the "or add it to ${ENV_FILE} first" tip is now accurate.

scripts/e2e.sh
ret2libc: Resolved by removing `--task-json` entirely in d8ee169 — part of a broader simplification (the script now only exposes `--budget`, `--task-duration`, `--image-tag`, `--no-pull`), so the silent-fallback path no longer exists.

scripts/e2e.sh
ret2libc: Fixed in d8ee169. The response now goes to a `mktemp` file tracked in `$TASK_RESP` and removed in `on_exit` (trap covers EXIT/INT/TERM), so the predictable name and the SIGINT/SIGTERM leak are both gone.

scripts/e2e.sh
ret2libc: Removed in d8ee169 — both the `--no-build` case branch and its usage line are gone (along with `--keep-up`/`--skip-wait`/`--sarif`/`--task-json`/per-phase `--*-timeout`), so the phantom "deprecated alias" framing no longer exists.

scripts/e2e.sh
```diff
@@ -0,0 +1,480 @@
...
+# Approve the patch (the local UI requires explicit approval, unlike scored
+# rounds where it is automatic). competition_patch_id= only appears once the
+# scheduler has built+verified+submitted the patch, well after the patch was
+# generated, so poll for it rather than capturing once (which always races).
+PATCH_LINE="$(wait_capture scheduler 'competition_patch_id=[0-9a-fA-F-]' \
+    "$MILESTONE_TIMEOUT" "competition_patch_id (for approval)" || true)"
+if [[ -n "$PATCH_LINE" ]]; then
+    PATCH_ID=$(printf '%s' "$PATCH_LINE" | sed -n 's/.*competition_patch_id=\([^ ]*\).*/\1/p')
+    # Task id is inside the first [...] block, after the last ':'.
+    TASK_ID=$(printf '%s' "$PATCH_LINE" | sed -n 's/.*\[\([^]]*\)\].*/\1/p' | sed 's/^[^:]*://')
```
hbrodin, line 422: The inline comment on the previous line says "Task id is inside the **first** `[...]` block, after the **last** `:`" — but the regex does the opposite of both: `s/.*\[\([^]]*\)\]/\1/` uses greedy `.*` and so captures the **last** `[...]` block, and `s/^[^:]*://` strips up to the **first** `:`.

It works today because `log_entry()` in `orchestrator/src/buttercup/orchestrator/scheduler/submissions.py` emits exactly one bracketed `[{idx_msg}{task_id}]` per line, and `task_id` contains no `:`. If anyone formats another bracketed token into the log message (e.g. appends bracketed context, or adds a field to `MaxLengthFormatter`'s format string in `common/logger.py`), this silently captures the wrong block; the subsequent `curl .../v1/task/${TASK_ID}/patch/${PATCH_ID}/approve` would then return 404 → `patch-approve: HTTP fail` with no diagnostic pointing back to the parse.

Repro: `printf '%s\n' '… [2:taskid] … extra [foo:bar]' | sed -n 's/.*\[\([^]]*\)\].*/\1/p' | sed 's/^[^:]*://'` yields `bar` instead of `taskid`.

Either tighten the regex (e.g. match a single `\[[^:]+:[^]]+\]` rather than `.*\[…\]`) or update the comment to match what the code actually does.

hbrodin (approved):
LGTM. One inline note on TASK_ID parsing — fragility, not a current bug.

PR #553, opened 2026-05-15
feat(llm): bump seed-gen/patcher to newest models + fix GPT-5.x temperature

ret2libc:
Stacked on **#544** (`worktree-add-newer-models`), which adds the new model definitions (`claude-4.6-sonnet`, `openai-gpt-5.5`, `openai-gpt-5.4-mini`, …). Base is #544's branch; merge after #544 (or retarget to `main` once #544 lands).

## 1. Bump seed-gen and patcher agents to the newest models

Tier-preserving upgrade onto the models from #544:

| Component | Before | After |
|---|---|---|
| seed-gen (all tasks) | primary `claude-4.5-sonnet`; fallbacks `claude-4-sonnet, gpt-4.1, gemini-pro` | primary `claude-4.6-sonnet`; fallbacks `openai-gpt-5.5, gemini-pro` |
| patcher SWE / RootCause / Reflection | primary `openai-gpt-4.1`; fallback `claude-4.5-sonnet, gemini-pro` | primary `openai-gpt-5.5`; fallback `claude-4.6-sonnet, gemini-pro` |
| patcher ContextRetriever | `openai-gpt-4.1`; cheap `openai-gpt-4.1-mini` | `openai-gpt-5.5`; cheap `openai-gpt-5.4-mini` |
| patcher code-snippet / test-instr helpers | `openai-gpt-4.1` / `claude-4.5-sonnet` | `openai-gpt-5.5` / `claude-4.6-sonnet` |

`gemini-pro` left as-is (not part of #544). The redundant `claude-4-sonnet` seed-gen fallback was dropped since `claude-4.6-sonnet` is now primary.

## 2. Fix: drop unsupported `temperature` for GPT-5.x in litellm

GPT-5.x reasoning models reject any non-default `temperature` with HTTP 400 (`Unsupported value: 'temperature' ... Only the default (1) value is supported`). The common LLM helpers send `temperature=0.1/0.3`, so putting patcher on `openai-gpt-5.5` broke every patcher LLM call.

A global `litellm_settings: drop_params: true` is **not sufficient**: the pinned `litellm 1.57.8` doesn't recognize these new model names, so it doesn't know `temperature` is unsupported and forwards it anyway. Fix adds an explicit, version-independent `additional_drop_params: ["temperature"]` to each `openai-gpt-5.x` entry, plus the global `drop_params: true` for other reasoning models litellm does recognize.

## Validation

Docker e2e against `example-libpng` (low LLM budget):
- **seed-gen** on `claude-4.6-sonnet`: 200+ healthy LLM turns, generated seeds, 0 errors.
- **patcher** on `openai-gpt-5.5`: agent loop runs with **0** temperature-400 / `unsupported_value` / litellm rejection errors (consistently failed before the litellm fix).


litellm/litellm_config.yaml
```diff
@@ -89,6 +89,32 @@ model_list:
       tpm: 150000000
       rpm: 30000
 
+  - model_name: openai-gpt-5.4-mini
+    litellm_params:
+      model: openai/gpt-5.4-mini
+      api_key: os.environ/OPENAI_API_KEY
+      # GPT-5.x reasoning models only accept temperature=1; this pinned litellm
+      # (1.57.8) doesn't know these new names, so drop temperature explicitly.
+      additional_drop_params: ["temperature"]
```
hbrodin, line 98: The same parallel `proxy_config` block in `deployment/k8s/values.yaml` (lines 366, 373, 380 for the three gpt-5.x entries, plus the `general_settings` section at ~452) doesn't get these edits. Per the comment at `values.yaml:273` it's a "Direct inclusion of the litellm_config.yaml content" — a manually-synced copy, and commit `4937a76` updated both files together when these models were introduced. As-is the Docker compose path is fixed but Helm deployments still send `temperature` to gpt-5.x and hit the same HTTP 400 the PR description describes.

Two options:

1. Mirror the four edits into `values.yaml`: add `additional_drop_params: ["temperature"]` to each `openai-gpt-5.x` entry, and add a `litellm_settings: drop_params: true` block next to `general_settings`.
2. Eliminate the duplication — the litellm-helm chart accepts `proxy_config` as a value, so the dev YAML can be the single source via `.Files.Get "litellm/litellm_config.yaml"` or a small generator step. Worth a follow-up regardless of which path you pick here.

patcher/src/buttercup/patcher/agents/context_retriever.py
```diff
@@ -587,8 +587,8 @@ def sh(
 def _are_test_instructions_valid(instructions: str, output: bytes, error: bytes) -> bool:
     """Validate a set of test instructions by executing them inside the project environment."""
     llm = create_default_llm(
-        model_name=ButtercupLLM.OPENAI_GPT_4_1.value,
-        fallback_models=[ButtercupLLM.CLAUDE_4_5_SONNET, ButtercupLLM.GEMINI_PRO],
+        model_name=ButtercupLLM.OPENAI_GPT_5_5.value,
```
hbrodin, line 590: `_are_test_instructions_valid` is a yes/no validator, and `create_default_llm` forwards the helper's default `temperature=0.1` (`common/llm.py:110`). After this PR the proxy strips `temperature` for gpt-5.x, so the primary silently runs at the model default (~1.0) while the Claude/Gemini fallbacks still get 0.1. For a binary validator that's a real consistency hazard — repeated calls on identical input can flip.

`rootcause.py:128-135` and `swe.py:345-352` already pass `"temperature": 1` explicitly in `kwargs` for their gpt-5.5 targets. Doing the same here, and at `common.py:437` (`_create_understand_code_snippet_chain`), makes the behavior local-readable instead of depending on proxy-side stripping:

```python
llm = create_default_llm(
    model_name=ButtercupLLM.OPENAI_GPT_5_5.value,
    temperature=1,
    fallback_models=[ButtercupLLM.CLAUDE_4_6_SONNET, ButtercupLLM.GEMINI_PRO],
)
```

Since this call is a binary predicate, an alternative is flipping the primary to `CLAUDE_4_6_SONNET` (which honors low temperature) so the verdict stays deterministic regardless of which model serves the request.

PR #554, opened 2026-05-18
fix(program-model): drop redundant deadsnakes PPA on Ubuntu 24.04

ret2libc:
Ubuntu 24.04 (noble) ships `python3.12` in the default archive, so the deadsnakes PPA / `software-properties-common` in `program-model/Dockerfile` only added an external launchpad.net dependency for no benefit (seed-gen/patcher already install python3.12 from the default archive). This drops it, matching the other component Dockerfiles and removing an unnecessary network dependency that can fail builds on networks without launchpad access.

### DIFF
--- a/program-model/Dockerfile
+++ b/program-model/Dockerfile
@@ -2,8 +2,6 @@ ARG BASE_IMAGE=ubuntu:24.04
 
 # hadolint ignore=DL3006
 FROM $BASE_IMAGE AS base
-# Ubuntu 24.04 (noble) ships python3.12 in the default archive, so the
-# deadsnakes PPA / software-properties-common is unnecessary here.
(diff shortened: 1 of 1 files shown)


evandowning (approved):
LGTM

program-model/Dockerfile
```diff
@@ -2,10 +2,9 @@ ARG BASE_IMAGE=ubuntu:24.04
 
 # hadolint ignore=DL3006
 FROM $BASE_IMAGE AS base
+# Ubuntu 24.04 (noble) ships python3.12 in the default archive, so the
+# deadsnakes PPA / software-properties-common is unnecessary here.
```
ret2libc:
```suggestion
```

hbrodin (approved):
LGTM

